PR #1, opened 2018-07-06
Support github URLs with non-standard filenames

jonludlam:
Github download URLs are normally of the form:

    https://github.com/proj/repo/archive/tag.extension

but it also supports URLs of the form:

    https://github.com/proj/repo/archive/tag/filname.extension

which allows better control of the final filename. This patch
adds support for the latter.

Signed-off-by: Jon Ludlam <jonathan.ludlam@citrix.com>


avsm:
thanks!

PR #2, opened 2018-10-03
git+ssh support

UnrealAkama:
Just moving an internal project to duniverse and needed these changes for support. This supports adding a opam repo to the duniverse switch and then pulling git+ssh packages from it when it's on a private git server.

### DIFF
--- a/src/opam_cmd.ml
+++ b/src/opam_cmd.ml
@@ -63,6 +63,10 @@ let tag_from_archive archive =
   in
   match Uri.scheme uri with
   | Some "git+http" | Some "git+https" -> Some "master"
+  | Some "git+ssh" -> (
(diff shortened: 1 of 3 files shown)


src/opam_cmd.ml
```diff
@@ -126,7 +130,12 @@ let classify_package ~package ~dev_repo ~archive ~pins () =
                   let repo = strip_ext repo in
                   (`Github (user, repo), tag)
               | tl -> err "wierd github url" )
-            | Some host -> (`Unknown host, tag)
+            | Some host -> (
+              match String.is_prefix "git" archive with
```
avsm: I think you can use `Uri.scheme` here, but can be done in a future patch

avsm:
thanks!

PR #4, opened 2019-04-08
Move vendored sources from ocaml_modules to duniverse

NathanReb:
The name `ocaml_modules` might be slightly confusing as we're not actually vendoring ocaml modules but entire dune projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ DESCRIPTION
 
        It works by analysing opam package metadata and calculating a set of
        git tags that can be cloned into the local repository into an
-       ocaml_modules subdirectory. Once the external code has been pulled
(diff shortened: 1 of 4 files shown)


NathanReb:
Might also be worth renaming `.duniverse` or at least renaming `duniverse_dir` to `dot_duniverse_dir` in `config.ml` so that `duniverse` developers don't get confused.

Not sure what's wrong with the CI but I'll look into it!

avsm:
lgtm! I think that we need to remove the .duniverse dir in favour of a single `dune-get` sexp file that has all the metadata, so no need for that renaming -- just unifying the metadata would be good instead so that there's no more need for a _duniverse directory at all.

NathanReb:
Ok perfect, sounds good to me as well!

PR #5, opened 2019-04-10
Use an opam repo to resolve dune overlays

Julow:
Use an opam repository to resolve dune overlay packages, instead of hardcoding the list in the source code.

The repo is here: https://github.com/dune-universe/opam-overlays

The repository is added to the local switch with the name `dune-overlays`, adding potential name collision.

This adds a regression: If there is no dune overlay for the lastest version of a package, the upstream version will be selected.
It can be fixed by checking if a dunified older version exists in the "package does not depend on dune" branch.

### DIFF
--- a/src/opam_cmd.ml
+++ b/src/opam_cmd.ml
@@ -62,8 +62,8 @@ let tag_from_archive archive =
     List.rev path |> List.hd |> tag_of_file ?prefix
   in
   match Uri.scheme uri with
-  | Some "git+http" | Some "git+https" -> Some "master"
-  | Some "git+ssh" -> (
+  | Some "git+http" | Some "git+https"
+  | Some "git+ssh" | Some "git" -> (
     match String.cuts ~empty:false ~sep:"#" archive with
     | [repo; tag] -> Some tag
     | _ -> Some "master")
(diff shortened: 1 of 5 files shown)


src/opam_cmd.ml
```diff
@@ -62,8 +62,8 @@ let tag_from_archive archive =
     List.rev path |> List.hd |> tag_of_file ?prefix
   in
   match Uri.scheme uri with
-  | Some "git+http" | Some "git+https" -> Some "master"
-  | Some "git+ssh" -> (
+  | Some "git+http" | Some "git+https"
```
Julow: I made an unrelated change here. Also, why are github urls treated specially here?

emillon: It's to determine the tag used for that release. For github URLs, it's possible to write a URL -> tag function, it's not that easy for other URLs.

NathanReb: That's actually not entirely true, at this point the tag has already been determined using `tag_from_archive` (which doesn't necessarily rely on `archive` being a github URL) so we should be able to make it work with other URLs.

emillon: I'll excuse myself, then.

:arrow_right:

NathanReb: Hmm ok forget about my comment, I mistook this pattern-matching with the one in `classify_package` so my bad!

NathanReb: But this functions has indeed some weird behaviour and the github case is actually never triggered for `git+https` which seems weird.

This is showcased in the tests I added in #6

PR #6, opened 2019-04-10
Extract duniverse_lib and add tests

NathanReb:
This PR splits `duniverse` code between a private library that lives under `lib/` and the actual CLI tool that lives under `bin/`.

While doing so I took the liberty to fix some of the warnings that were there and that are picked up by dune's default profile.

I also added some tests. For now only `Opam_cmd.tag_from_archive` is tested but it sets a ground to add more tests in the future.

Regarding the tests you'll note that they actually are characterization tests and I'm not sure the behaviour they outline is the expected one, especially in the `"git+https://a.com/user/repo#v0.0.0"` case but at least now we have a reference and we can discuss it!

Let me know what you think!


NathanReb:
Hmm in fact the weird `tag_from_archive` behaviour is fixed in #5 !

NathanReb:
Just rebased on top of #7 to get the :heavy_check_mark:

Makefile
```diff
@@ -9,6 +9,9 @@ clean:
 doc:
 	dune build @doc --profile=release
 
+unit:
```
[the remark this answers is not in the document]
avsm:
```suggestion
test:
```

agreed, `test` is more conventional here

bin/dune
```diff
@@ -0,0 +1,12 @@
+(executable
+ (name duniverse)
+ (public_name duniverse)
+ (libraries
+    cmdliner
+    duniverse_lib
```
avsm: I'd prefer to call this library just `duniverse`, to follow the other ones.  The CLI driver can just be `main.ml` or `cli.ml` and then installed as `duniverse` via `(public_name duniverse)` in the executable stanza

test/test_opam_cmd.ml
```diff
@@ -0,0 +1,28 @@
...
+  ; make_test ~archive:"https://github.com/user/repo/archive/v0.0.0/archive.tbz" ~expected:(Some "v0.0.0")
+  ; make_test ~archive:"https://ocaml.janestreet.com/ocaml-core/4.07.1/files/package-v0.0.0.tbz" ~expected:(Some "v0.0.0")
+  ; make_test ~archive:"https://ocaml.janestreet.com/janestreet/repo/releases/download/v0.0.0/file.tbz" ~expected:(Some "v0.0.0")
+  ; make_test ~archive:"https://ocaml.janestreet.com/janestreet/repo/archive/v0.0.0.tbz" ~expected:(Some "v0.0.0")
+  ; make_test ~archive:"https://ocaml.janestreet.com/janestreet/malformed" ~expected:None
+  ; make_test ~archive:"https://gitlab.camlcity.org/some/path/file-v0.0.0.tbz" ~expected:(Some "v0.0.0")
+  ; make_test ~archive:"https://download.camlcity.org/some/path/file-v0.0.0.tbz" ~expected:(Some "v0.0.0")
+  ; make_test ~archive:"https://ocamlgraph.lri.fr/some/path/file-0.0.0.tbz" ~expected:(Some "v0.0.0")
+  ; make_test ~archive:"https://erratique.ch/some/path/file-0.0.0.tbz" ~expected:(Some "v0.0.0")
+  ; make_test ~archive:"https://other.domain.com/some/path/file-v0.0.0.tbz" ~expected:(Some "v0.0.0")
```
avsm: Most glad to see these tests :-)

bin/dune
NathanReb: Wouldn't it make things more confusing in the end?

Happy to change it anyway but I tend to avoid giving the same name to two different things so that new comers to the project don't have to go through the dune files to figure out what is compiled to what.

NathanReb:
Pushed a few changes after @emillon's feedback!

README.md
```diff
@@ -31,12 +31,12 @@ DESCRIPTION
        required to build a project using the dune build tool.
 
        It works by analysing opam package metadata and calculating a set of
-       git tags that can be cloned into the local repository into an
-       duniverse subdirectory. Once the external code has been pulled
-       into the repository, a single dune build command is sufficient to
-       build the whole project in a standalone fashion, without opam being
-       required. This is a particularly convenient way of publishing CLI
-       tools to users who do not need the full power of opam.
+       git tags that can be cloned into the local repository into a duniverse
```
avsm:
```suggestion
       git tags that can be cloned into the local repository into a `duniverse/`
```

bin/dune
avsm: It's more confusing to not be able to predict the findlib name from the opam package name, I find. Either way, it doesn't matter much so I'll merge this as-is and we can change later if we want.  The `duniverse_lib` probably shouldn't be installed in the default package since it's really an internal library for the driver -- there are no expected external consumers outside of the CLI tool to start with are there?

NathanReb: Not that I can think of right now.

It is indeed not installed atm!

PR #7, opened 2019-04-10
Fix CI script

NathanReb:
This is just a quick-fix, I'm not sure what the future of `duniverse` CI should be.

I first thought about using the regular `.travis-docker.sh`, which would work but I seem to recall a mention of `duniverse` being a standalone tool with vendored dependencies, in which case updating the custom CI script seems better!

### DIFF
--- a/.docker.sh
+++ b/.docker.sh
@@ -1,12 +1,12 @@
 #!/usr/bin/env bash
-DISTRO=${DISTRO:-alpine-3.7}
+DISTRO=${DISTRO:-alpine-3.9}
 
 set -ex
 # TODO opam2 depext
 case $DISTRO in
-alpine-*) sudo apk add m4 ;;
-debian-*) sudo apt update; sudo apt -y install m4 pkg-config ;;
-ubuntu-*) sudo apt update; sudo apt -y install m4 pkg-config ;;
+alpine*) sudo apk add m4 ;;
+debian*) sudo apt update; sudo apt -y install m4 pkg-config ;;
+ubuntu*) sudo apt update; sudo apt -y install m4 pkg-config ;;
 esac
 
 sudo chown -R opam /home/opam/src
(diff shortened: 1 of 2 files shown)


NathanReb:
It seems that the `alpine-3.7` build is broken because the `uri` version picked by `opam` has a bad checksum.

Running `opam update` doesn't solve the issue since the docker image uses a pinned repo (I wasn't aware of that!), which is out-of-date for the alpine-3.7 image.

I took the liberty to update some of the images to use the latest versions, ie `debian-stable`, `alpine`, and `ubuntu`. I kept the `ubuntu-16.04` as it seemed to be tested on purpose.

.docker.sh
```diff
@@ -1,16 +1,18 @@
...
+DISTRO=${DISTRO:-alpine-3.9}
 
 set -ex
 # TODO opam2 depext
 case $DISTRO in
-alpine-*) sudo apk add m4 ;;
-debian-*) sudo apt update; sudo apt -y install m4 pkg-config ;;
-ubuntu-*) sudo apt update; sudo apt -y install m4 pkg-config ;;
+alpine*) sudo apk add m4 ;;
+debian*) sudo apt update; sudo apt -y install m4 pkg-config ;;
```
Julow: Shorter: `debian*|ubuntu*))`

.docker.sh
```diff
@@ -1,16 +1,18 @@
...
+ubuntu*) sudo apt update; sudo apt -y install m4 pkg-config ;;
 esac
 
 sudo chown -R opam /home/opam/src
 cd /home/opam/src
-opam install -y dune 
+opam pin add --no-action duniverse .
+opam update
+opam install --deps-only -t duniverse
 cd /home/opam/src
```
Julow: This line is not necessary.

avsm:
> It seems that the alpine-3.7 build is broken because the uri version picked by opam has a bad checksum.

Yeah, it's now better to just use `alpine` as an alias for the latest build of Alpine, since specific OS releases will lag behind after they drop out of the support cycle.

PR #8, opened 2019-04-12
Use a new and temporary opam state for each duniverse-opam run

NathanReb:
This PR remove the local switch from `.duniverse` and uses a completely new `.opam` for each `duniverse opam` run instead to make this step stateless.

It comes at a price since subsequent run of `duniverse opam` won't benefit from the switch creation.

I rebased this on top of #10  (and thus set it to draft) to get the working CI. Only the last commit is relevant to this PR!

### DIFF
--- a/lib/opam_cmd.ml
+++ b/lib/opam_cmd.ml
@@ -135,17 +135,17 @@ let classify_package ~package ~dev_repo ~archive ~pins () =
             | false -> (`Unknown host, tag) )
           | None -> err "dev-repo without host" ) )
 
-let check_if_dune ~repo package =
-  Exec.get_opam_depends ~repo (string_of_package package)
+let check_if_dune ~root package =
+  Exec.get_opam_depends ~root (string_of_package package)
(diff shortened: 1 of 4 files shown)


lib/opam_cmd.ml
```diff
@@ -263,21 +263,31 @@ let calculate_duniverse ~repo file =
...
+  in
+  Exec.init_opam_and_remotes ~root ~compiler ~remotes:(dune_overlays::user_specified_remotes) ()
+
+let init_duniverse repo branch roots excludes pins compiler remotes () =
   Logs.app (fun l ->
       l "%aCalculating Duniverse on the %a branch." pp_header header
         Fmt.(styled `Cyan string)
         branch ) ;
   let file = Fpath.(repo // Config.opam_lockfile) in
+  Bos.OS.Dir.tmp ".duniverse-opam-root-%s"
```
avsm: nice, and this takes care of cleaning up at program exit as well

PR #9, opened 2019-04-15
Faster duniverse opam fetch using one single command.

TheLortex:
Instead of calling opam three times per package dependency, fetch and parse everything at once.
This is _a lot_ faster.

### DIFF
--- a/lib/exec.ml
+++ b/lib/exec.ml
@@ -136,7 +136,7 @@ let opam_switch_create_empty ~root () =
   let cmd = opam_cmd ~root "switch" % "create" % "empty" % "--empty" % "--no-install" in
   run_and_log cmd
 
-let get_opam_fields_lines ~root fields packages =
+let run_opam_show ~root ~fields ~packages =
   let fields = List.map (fun s -> s^":") fields in
   let fields = "name"::fields in
   let fields_str = String.concat ~sep:"," fields
@@ -149,39 +149,6 @@ let get_opam_fields_lines ~root fields packages =
     % "-f" % fields_str %% (of_list packages_str)
   in
   run_and_log_l cmd
-
-let parse_opam_show_line line = match String.cut ~sep:" " line with
-  | None -> "",""
-  | Some (field, value) -> (String.trim field, String.trim value)
-
-module Opam_show_result = struct
-  type t = OpamParserTypes.value Types.StrMap.t
-
-  let make ~root fields packages =
-    get_opam_fields_lines ~root fields packages
-    >>= fun lines ->
-    let rec parse cur_name map = function
-    | [] -> Ok map
-    | line::next -> (
-      match parse_opam_show_line line with
-      | ("name", pkg_name) -> parse pkg_name map next
-      | (field, value_str) ->
-        (match value_str with
-        | "" -> Ok OpamParserTypes.(List (("", 0, 0), []))
-        | r -> (
-          try Ok (OpamParser.value_from_string r "") with OpamLexer.Error _ ->
-            Error (`Msg (Fmt.strf "parsing error for: '%s'" r)) ))
-        >>= fun value ->
-        let field = String.trim ~drop:(fun x -> x == ':') field in
-        let map = StrMap.add (cur_name^"."^field) value map in
-        parse cur_name map next
-    )
-    in
-    parse "" Types.StrMap.empty lines
-
-  let get ~package ~field t = Types.StrMap.find_opt (package^"."^field) t
-end
-
 
 let opam_add_remote ~root { Types.Opam.Remote.name; url } =
   let open Cmd in
--- a/lib/exec.ml
+++ b/lib/exec.ml
@@ -137,14 +137,12 @@ let opam_switch_create_empty ~root () =
   run_and_log cmd
 
 let run_opam_show ~root ~fields ~packages =
-  let fields_str = String.concat ~sep:"," fields
-  in
-  let packages_str = List.map Types.Opam.string_of_package packages
-  in
+  let fields_str = String.concat ~sep:"," fields in
+  let packages_str = List.map Types.Opam.string_of_package packages in
   let cmd =
     let open Cmd in
-    opam_cmd ~root "show" %"--color=never" % "--normalise"
-    % "-f" % fields_str %% (of_list packages_str)
+    opam_cmd ~root "show" % "--color=never" % "--normalise" % "-f" % fields_str
+    %% of_list packages_str
   in
   run_and_log_l cmd
 
--- a/lib/exec.ml
+++ b/lib/exec.ml
@@ -126,57 +126,6 @@ let run_opam_package_deps ~root packages =
   in
   run_and_log_l cmd
 
-let get_opam_field ~root ~field package =
-  let field = field ^ ":" in
-  let cmd =
-    let open Cmd in
-    opam_cmd ~root "show" % "--color=never" % "--normalise" % "-f" % field % package
-  in
-  run_and_log_s cmd >>= fun r ->
-  match r with
-  | "" -> Ok OpamParserTypes.(List (("", 0, 0), []))
-  | r -> (
-    try Ok (OpamParser.value_from_string r "")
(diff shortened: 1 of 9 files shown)


src/exec.ml
```diff
@@ -180,63 +180,42 @@ let run_opam_package_deps ~repo packages =
...
+  | [] -> Ok map
+  | line::next -> (
+    match String.cut ~sep:" " line with
+    | None -> parse cur_name map next
+    | Some ("name", value) -> parse (String.trim value) map next
+    | Some (field, value_str) ->
+      (match (String.trim value_str) with
+      | "" -> Ok OpamParserTypes.(List (("", 0, 0), []))
+      | r -> (
+        try Ok (OpamParser.value_from_string r "") with _ ->
```
NathanReb: I think it would be better to only catch the relevant exception here, probably something like:
```suggestion
        try Ok (OpamParser.value_from_string r "") with OpamLexer.Error _ ->
```

src/opam_cmd.ml
```diff
@@ -135,30 +135,87 @@ let classify_package ~package ~dev_repo ~archive ~pins () =
...
+
+
+let get_opam_info ~repo ~pins packages =
+  let fields = ["dev-repo"; "url.src"; "depends"] in
+  Exec.get_opam_fields ~repo fields packages
+   >>= fun data ->
+  let packages = List.map
+    (fun pkg ->
+      let name = pkg.name in
+      let archive_url_line = Types.StrMap.find_opt (name^".url.src") data in
```
NathanReb: What would you think about extracting an abstract type for those `opam show` results that you could query with a more suited API, something in the lines of:
```ocaml
let archive_url_line = Opam_show_result.get ~package:pkg.name ~field:"url.src" data in
```
?

NathanReb:
I think this is the kind of change that showcases how much of an improvement it would be to use the opam library instead of the binary!

I'll look futher into that!

avsm:
I like this change, but it does indeed need to rebased to work with the tmpdir switches (now merged) as they are important to keep the tool stateless.

NathanReb:
There might be something to do by modifying init_opam to run:

opam init --bare --no-setup
opam switch create <compiler> --empty

NathanReb:
I'm currently working on a PR to use an empty switch and thus avoid the lengthy build of a compiler but that requires some care as the `opam list` invocation needs to be filter since it will always return `base-bigarray`, `base-thread`, `ocaml-config`, `ocaml-base-compiler` and so on.

If you're patient enough to wait for this, rebasing your PR should be easier as only the `opam show` improvements will be relevant then!

src/exec.ml
```diff
@@ -180,63 +180,42 @@ let run_opam_package_deps ~repo packages =
...
-        (Fmt.strf
-           "Unable to parse opam depends for %s\n\
-            Try `opam show --normalise -f depends: %s` manually"
-           package package)
+  run_and_log_l cmd
+
+let get_opam_fields ~repo fields packages =
+  get_opam_fields_lines ~repo fields packages
+  >>= fun lines ->
+  let rec parse cur_name map = function
```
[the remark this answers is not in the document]
TheLortex: I agree, this one is quite a chunk

src/opam_cmd.ml
TheLortex: Indeed that would be more readable

TheLortex:
I can wait for your changes, I'll just update the PR accoding to your comments !

TheLortex:
Can we have a review on this ?

lib/exec.mli
```diff
@@ -52,22 +52,21 @@ val git_merge :
...
     resolves the transitive dependencies of [packages]. *)
 
+(** [run_opam_show ~root ~fields ~packages] runs opam show to get [fields] for each package in [packages].*)
+val run_opam_show :
+  root:Fpath.t ->
+  fields:string list ->
+  packages:Types.Opam.package list ->
+  (string list, [> Rresult.R.msg ]) result
+
+(** [init_local_opam_switch ~opam_switch ~root ~remotes ()] creates a fresh opam state and switch
```
Julow: Is this a rebase fail ?

lib/exec.ml
```diff
@@ -187,6 +136,20 @@ let opam_switch_create_empty ~root () =
   let cmd = opam_cmd ~root "switch" % "create" % "empty" % "--empty" % "--no-install" in
   run_and_log cmd
 
+let run_opam_show ~root ~fields ~packages =
+  let fields = List.map (fun s -> s^":") fields in
```
Julow: You should not concatenate `":"` here. The colon is part of the name of field and some fields do not have it.

lib/opam_cmd.ml
```diff
@@ -125,27 +125,100 @@ let classify_package ~package ~dev_repo ~archive ~pins () =
...
+  | "" -> Ok None
+  | uri -> Ok (Some uri)
+
+let parse_opam_depends ~package v =
+  let open OpamParserTypes in
+  match extract_opam_value v with
+  | Some (Ok (List (_, vs))) ->
+      let ss =
+        List.fold_left
+          (fun acc -> function String (_, v) -> v :: acc
```
Julow: Could be merged with the next case.

lib/opam_cmd.ml
```diff
@@ -125,27 +125,100 @@ let classify_package ~package ~dev_repo ~archive ~pins () =
...
+    in
+    Some res
 
-let get_opam_info ~root ~pins package =
-  Exec.get_opam_dev_repo ~root (string_of_package package) >>= fun dev_repo ->
-  Exec.get_opam_archive_url ~root (string_of_package package) >>= fun archive ->
-  let dev_repo, tag = classify_package ~package ~dev_repo ~archive ~pins () in
-  check_if_dune ~root package >>= fun is_dune ->
-  Logs.info (fun l ->
+let parse_string ~field ~package v =
```
Julow: Why not move this to `Opam_show_result` ? (Because of the dependency on `OpamParser` ?)

lib/exec.mli
TheLortex: Yup, fixed that.

lib/exec.ml
TheLortex: I agree, so now `:` is not treated specially, in order to avoid surprises

lib/opam_cmd.ml
TheLortex: `Opam_show_result` doesn't need to know how fields should be parsed so I don't think it's necessary to move this.

NathanReb: The main reason why we did this was to avoid carrying dummy OpamValue with irrelevant locations around. That made testing harder and indeed, as @TheLortex said, `Opam_show_result` is just an abstraction above the `opam show` output, it's not strictly required to parse everything in there.

Julow: I understand.
What about moving this functions as is ? Also, it could call the `get` function as it has the `package` and `field` values.

lib/exec.ml
Julow: You still pass field names without colon to `parse_string` but that's just for error messages.

NathanReb:
@Julow @TheLortex is there anything still blocking this?

TheLortex:
I don't think so !

lib/opam_cmd.ml
```diff
@@ -125,27 +125,89 @@ let classify_package ~package ~dev_repo ~archive ~pins () =
...
+
+let get_opam_info ~root ~pins packages =
+  let fields = [ "name"; "dev-repo:"; "url.src:"; "depends:" ] in
+  Exec.run_opam_show ~root ~packages ~fields >>= fun lines ->
+  Opam_show_result.make lines >>= fun data ->
+  let packages =
+    List.map
+      (fun pkg ->
+        let name = pkg.name in
+        let archive_url_line = Opam_show_result.get ~package:name ~field:"url.src:" data in
```
Julow: I suggest moving calls to `Opam_show_result.get` to `extract_opam_value`. The `package` and `field` arguments are needed in the `parse_*` functions and are passed in different ways.

TheLortex: Done. Indeed it improves readability.

PR #10, opened 2019-04-15
Add an .mli to the Exec module

NathanReb:
Depends on #6 

In an effort to clean up the `duniverse` repo I started adding `.mli` files for the duniverse internal library. `Exec` is the first in that series but I'm planning on doing so for the remaining modules.

I exposed the functions that were used outside the module and added a bit of documentation.

I took the liberty of getting rid of the unused values. Some of them were commented as so so I'm guessing that we're planning on using them at some point in the future. I removed them but I'm happy ot open an issue to document that soon to be feature if you feel like it's useful!

### DIFF
--- a/lib/exec.ml
+++ b/lib/exec.ml
@@ -24,8 +24,6 @@ let rec iter fn l =
 
 let err_log = OS.Cmd.err_file ~append:true Config.duniverse_log
 
-let out_log = OS.Cmd.(to_null)
-
 let run_and_log_s ?(ignore_error = false) cmd =
   OS.File.tmp "duniverse-run-%s.stderr"
   >>= fun tmp_file ->
@@ -125,8 +123,6 @@ let git_checkout_or_branch ~repo branch =
   | Ok () -> Ok ()
   | Error (`Msg _) -> git_checkout ~args:(Cmd.v "-b") ~repo branch
 
-let git_rm_rf ~repo file = run_git ~repo Cmd.(v "rm" % "-rf" % p file)
-
 let git_add_and_commit ~repo ~message files =
   run_git ~ignore_error:true ~repo Cmd.(v "add" %% files)
   >>= fun () ->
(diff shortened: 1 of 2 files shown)


bin/dune
```diff
@@ -0,0 +1,12 @@
+(executable
+ (name duniverse)
```
emillon: it's already in `public_name`:

```suggestion
```

dune-project
```diff
@@ -1,2 +1,2 @@
-(lang dune 1.0)
+(lang dune 1.2)
```
emillon: this requires a version bound in `opam`

lib/dune
```diff
@@ -0,0 +1,19 @@
...
+    fmt
+    logs
+    opam-core
+    opam-file-format
+    rresult
+    sexplib
+    uri
+  )
+  (preprocess
+    (per_module
```
emillon: this was in the original dune file, but I'd say not to bother with this optimization, this can be confusing when adding a converter in another place

lib/dune_cmd.ml
```diff
@@ -133,7 +133,7 @@ let gen_dune_lock repo () =
         ofile ) ;
   Ok ()
 
-let status repo target_branch () = Ok ()
+let status _repo _target_branch () = Ok ()
```
emillon: would it make sense to remove the params and clean the call sites?

lib/opam_cmd.ml
```diff
@@ -126,9 +126,9 @@ let classify_package ~package ~dev_repo ~archive ~pins () =
             | [user; repo] ->
                 let repo = strip_ext repo in
                 (`Github (user, repo), tag)
-            | tl -> err "wierd github url" )
+            | _ -> err "wierd github url" )
```
emillon:
```suggestion
            | _ -> err "weird github url" )
```

test/test_opam_cmd.ml
```diff
@@ -0,0 +1,28 @@
...
+    let test_name = "tag_from_archive: " ^ archive in
+    let test_fun () =
+      let actual = Duniverse_lib.Opam_cmd.tag_from_archive archive in 
+      Alcotest.(check (option string)) archive expected actual
+    in
+    (test_name, `Quick, test_fun)
+  in
+  [ make_test ~archive:"git+http://a.com/user/repo" ~expected:(Some "master")
+  ; make_test ~archive:"git+https://a.com/user/repo" ~expected:(Some "master")
+  ; make_test ~archive:"git+https://a.com/user/repo#v0.0.0" ~expected:(Some "v0.0.0")
```
emillon: what do you think about using `1.2.3` for these version numbers? 0.0.0 seems a bit too symmetric.

Makefile
```diff
@@ -9,6 +9,9 @@ clean:
 doc:
 	dune build @doc --profile=release
 
+test:
+	dune runtest
```
emillon: it depends a bit one where it's going to be invoked from, but you might want `--profile release` here

NathanReb: It's meant for CI so I'd rather leave it as it is.

NathanReb:
I adressed your suggestions in the relevant PR: #6

avsm:
> Some of them were commented as so so I'm guessing that we're planning on using them at some point in the future.

The opposite -- the automatic git management just didn't work that well, so they were on their way out. Thanks for the cleanup!

PR #11, opened 2019-04-16
Fix a few quirks in the documentation and the generated README

NathanReb:
We accidentally modified the README which is generated from the `Cmdliner` documentation of the `duniverse` binary and thus every time we build the project, this change was overwritten.

Might be worth moving this to a different file and having a proper README for the repo but that's a different story. I'll work on that as well!


avsm:
Woops, this is one of the downsides of the promote workflow :-)

PR #12, opened 2019-04-17
Use a temporary empty switch for dependencies resolution

NathanReb:
As a followup to #8 and #9 `duniverse opam` now uses an temporary opam state with an empty switch.

This makes it much faster as we don't have to compile ocaml anymore.
The output of `opam list --resolve` will now give us a bunch of new dependencies such as `ocaml-base-compiler` and `base-bigarray` but they are properly handled by `classify_package`. This was a bit cryptic and hard to figure out so we should probably improve that part of the code in the future but it works for now.

As a side effect, the `Opam_cmd.init_duniverse` function doesn't need the `compiler` argument anymore so I removed it, both from the command line arguments and from the `Types.Opam.t` type. Be aware that this is a breaking change to `duniverse opam` and `duniverse lock`.

Another point worth noting is that now, unless users specify an ocaml version in their `.opam` files, the solver will pick one by itself so we should probably encourage explicitly define one.

Let me know what you think!

### DIFF
--- a/lib/exec.mli
+++ b/lib/exec.mli
@@ -86,12 +86,10 @@ val get_opam_archive_url : root: Fpath.t -> string -> (string option, [> Rresult
     resolves the transitive dependencies of [packages]. *)
 val run_opam_package_deps : root: Fpath.t -> string list -> (string list, [> Rresult.R.msg]) result
 
-(** [init_local_opam_switch ~opam_switch ~root ~remotes ()] creates a fresh opam state and switch
-    with compiler [compiler] using [root] as OPAMROOT and adds the [remotes] opam repositories
-    to it. *)
+(** [init_local_opam_switch ~root ~remotes ()] creates a fresh opam state with a single
+    switch using [root] as OPAMROOT and adds the [remotes] opam repositories to it. *)
 val init_opam_and_remotes :
   root: Fpath.t ->
-  compiler: string ->
   remotes: Types.Opam.Remote.t list ->
   unit ->
   (unit, [> Rresult.R.msg]) result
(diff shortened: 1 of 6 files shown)


NathanReb:
CC @TheLortex

lib/exec.mli
```diff
@@ -86,12 +86,10 @@ val get_opam_archive_url : root: Fpath.t -> string -> (string option, [> Rresult
     resolves the transitive dependencies of [packages]. *)
 val run_opam_package_deps : root: Fpath.t -> string list -> (string list, [> Rresult.R.msg]) result
 
-(** [init_local_opam_switch ~opam_switch ~root ~remotes ()] creates a fresh opam state and switch
-    with compiler [compiler] using [root] as OPAMROOT and adds the [remotes] opam repositories
-    to it. *)
+(** [init_local_opam_switch ~opam_switch ~root ~remotes ()] creates a fresh opam state with a single
+    switch using [root] as OPAMROOT and adds the [remotes] opam repositories to it. *)
```
Julow: Comment should be `[init_opam_and_remotes ~root ~remotes ()]`.

NathanReb: Good catch!

NathanReb:
> For the version of the compiler, does it make sense to use the version in the user's path (ocamlc -version) by default ?

Yeah I think that could work and it would make a good default but if we do that we should expose an option to configure it as I can easily see that becoming annoying in some cases.

avsm:
Ooh this is good.  We haven't tackled the more general problem of how to make a duniverse compatible with _multiple_ ocaml versions yet.  So specifying a single version is a good idea, and the system compiler is a reasonable default in this case.  I dont think depending on the user PATH is quite as reproducible as just the system compiler...

PR #13, opened 2019-04-17
Add integration tests

NathanReb:
Depends on #12 on top of which it is rebased. This is required to have a fast enough CI as compiling ocaml in a docker was terribly slow.

Only the last commit is relevant to this PR, please don't merge before #12 !

This PR adds a simple integration test. We build a simple binary which depends both on a regular upstream packages and a dune-overlay one using `duniverse` and make sure everything work fine.

As I was working on this I figured we could probably improve our CI to make better use of the docker's cache but this can be done separately!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build clean test doc
+.PHONY: build clean test doc install
 
 build:
 	dune build 
@@ -11,3 +11,6 @@ doc:
 
 test:
 	dune runtest
+
+install:
+	dune install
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: build clean test doc
+.PHONY: build clean test doc install
```
Julow: I would be in favor of removing the `Makefile` and using `dune` directly.

avsm: I'd prefer to keep the makefile -- I use the convenience targets all the time :-)

NathanReb:
Rebased and the CI's green so I'm merging this since both of you approved! Thanks!

PR #14, opened 2019-04-17
Warn instead of failing when a package is invalid.

TheLortex:
Even if a duniverse fork is needed, the user might be able to fetch it manually so it doesn't really make sense to stop and not downloading anything.

### DIFF
--- a/lib/opam_cmd.ml
+++ b/lib/opam_cmd.ml
@@ -209,21 +209,6 @@ let get_opam_info ~root ~pins packages =
     (fun lst elem -> lst >>= fun lst -> elem >>| fun elem -> elem :: lst)
     (Ok []) packages
 
-let package_is_valid { package; dev_repo; _ } =
-  match dev_repo with
-  | `Error msg ->
-      Logs.warn (fun l -> l "Do not know how to deal with %a: %s" pp_package package msg);
-      false
-  | `Unknown msg ->
-      Logs.warn (fun l -> l "Need a Duniverse fork for %a: %s" pp_package package msg);
-      false
-  | _ -> true
-
-let check_packages_are_valid pkgs =
-  Logs.app (fun l ->
-      l "%aEliminating dependencies that are not understood by the Duniverse." pp_header header );
-  List.filter package_is_valid pkgs
-
 let filter_duniverse_packages ~excludes pkgs =
   Logs.app (fun l ->
       l "%aFiltering out packages that are irrelevant to the Duniverse." pp_header header );
@@ -258,9 +243,7 @@ let calculate_opam ~root ~root_packages ~excludes ~pins ~branch ~remotes =
   Logs.app (fun l ->
       l "%aQuerying local opam switch for their metadata and Dune compatibility." pp_header header
   );
-  get_opam_info ~root ~pins deps >>= fun packages ->
-  let packages = check_packages_are_valid packages in
-  filter_duniverse_packages ~excludes packages >>= fun packages ->
+  get_opam_info ~root ~pins deps >>= filter_duniverse_packages ~excludes >>= fun packages ->
   Ok { packages; root_packages; excludes; pins; branch; remotes }
 
 type packages_stats = { total : int; dune : int; not_dune : entry list }
--- a/lib/opam_cmd.ml
+++ b/lib/opam_cmd.ml
@@ -212,16 +212,17 @@ let get_opam_info ~root ~pins packages =
 let package_is_valid { package; dev_repo; _ } =
   match dev_repo with
   | `Error msg ->
-      R.error_msg (Fmt.strf "Do not know how to deal with %a: %s" pp_package package msg)
+      Logs.warn (fun l -> l "Do not know how to deal with %a: %s" pp_package package msg);
+      false
   | `Unknown msg ->
-      R.error_msg (Fmt.strf "Need a Duniverse fork for %a: %s" pp_package package msg)
-  | _ -> R.ok ()
+      Logs.warn (fun l -> l "Need a Duniverse fork for %a: %s" pp_package package msg);
+      false
+  | _ -> true
 
 let check_packages_are_valid pkgs =
   Logs.app (fun l ->
-      l "%aChecking that all dependencies are understood by the Duniverse." pp_header header );
-  let rec fn = function hd :: tl -> package_is_valid hd >>= fun () -> fn tl | [] -> R.ok () in
-  fn pkgs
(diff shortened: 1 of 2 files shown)


NathanReb:
I agree with @Julow  here. I think doing so would break the behaviour of duniverse after that so it shouldn't be the default. I agree this can be useful in development so we should indeed expose a CLI flag to allow that.

Since duniverse now relies on [opam-overlays](https://github.com/dune-universe/opam-overlays) I think we should encourage people that know of such forks to open a PR there.

Another solution to keep `duniverse` in line with the actual content of the `duniverse/` directory and files would be to allow you to run some kind of `duniverse install` which would let you specify what package you want and where you want to fetch its sources from. This is not as good as having an `opam-overlays` port but at least `duniverse` would explicitly of your dependency and where to fetch it.

I'm currently working on a new spec for duniverse which includes that.

avsm:
Out of curiosity, was there a particular failure mode that made you want this feature, @TheLortex?  What were you trying to do when this missing package happened?

i.e. if its a symptom that:

- upstream released a new version (like recently with fmt) and there is no upstream dune-universe overlay 
or...
- its a broken tag inference 
or...
- its an upstream opam repository issue

TheLortex:
Maybe an option would be nice, I don't know if the error is supposed to exist in the future. What I would like is to be able to ignore errors and download as many packages as possible, and fix them by hand if needed ! 
@avsm 
It fails for `ocamlfind`, as it's classified as `Unknown`

avsm:
Aha, I think we're missing an ocamlfind in the opam-overlays... https://github.com/dune-universe/lib-findlib  (this is to install the findlib ocaml libraries used by utop)

Julow:
I think that if it fails on `ocamlfind`, it's because an other non-dune package requires it. Dunifying `ocamlfind` won't help.

TheLortex:
I agree on the fact that it won't help. 
However something I find strange is that for `ocamlfind`, it fails with `Need a Duniverse fork for ocamlfind.1.8.0: gitlab.camlcity.org` whereas for other non-dune packages, such as `ocaml-freestanding`, it's only flagged as non-dune and shows the warning message `The bad news is that you will have to fork these to the Duniverse or port them upstream` which still allows to install dune-compatible packages.

avsm:
> I think that if it fails on ocamlfind, it's because an other non-dune package requires it. Dunifying ocamlfind won't help.

It will help, since there are dependencies on these libraries in, e.g. utop:

```
findlib             (version: 1.8.0)
findlib.dynload     (version: 1.8.0)
findlib.internal    (version: 1.8.0)
findlib.top         (version: 1.8.0)
```

TheLortex:
Rebased. Regarding the `best-effort` option, I don't think it's actually relevant. For most non-dune packages, duniverse already shows a warning message and installs as much as it can. It doesn't really make sense that an `Unknown` package should trigger a hard failure.

lib/opam_cmd.ml
```diff
@@ -212,16 +212,17 @@ let get_opam_info ~root ~pins packages =
 let package_is_valid { package; dev_repo; _ } =
   match dev_repo with
   | `Error msg ->
-      R.error_msg (Fmt.strf "Do not know how to deal with %a: %s" pp_package package msg)
+      Logs.warn (fun l -> l "Do not know how to deal with %a: %s" pp_package package msg);
```
NathanReb: How about moving the logs in a different function so that we only print logs about such packages if we explicit want to instead of it being a site effect of computing this `is_valid` property?

I.e. we would have a:
```ocaml
let log_invalid_packages packages =
  List.iter
   (function
      | {dev_repo = `Error msg; _} -> ...
      | {dev_repo = `Unknown host; _ } -> ...
      | _ ->  ())
   packages
```
That we would call at the appropriate time during the command execution.

lib/opam_cmd.ml
```diff
@@ -212,16 +212,17 @@ let get_opam_info ~root ~pins packages =
...
+      false
   | `Unknown msg ->
-      R.error_msg (Fmt.strf "Need a Duniverse fork for %a: %s" pp_package package msg)
-  | _ -> R.ok ()
+      Logs.warn (fun l -> l "Need a Duniverse fork for %a: %s" pp_package package msg);
+      false
+  | _ -> true
 
 let check_packages_are_valid pkgs =
```
NathanReb: This function should probably be renamed to `filter_invalid_packages` or `filter_unsupported_packages`.

NathanReb: Might actually be worth grouping all the package filtering together in a single function. That function could then be used in #23 to determine which package should be installed through `opam`.

lib/opam_cmd.ml
TheLortex: Done

lib/opam_cmd.ml
TheLortex: Renamed the function. Yes that would be a good refactor in the future.

TheLortex:
> I'd suggest we do one step at a time though and keep failing on `Error` packages for now.

Errors are "weird github url" and "dev-repo without host", so they are roughly the same kind of error as `Unknown`, I don't see why they should cause a hard failure.
 
> Actually I think the proper way to deal with them is to remove the `Error` variant from the dev_repo type and make `classify_package` return a `(dev_repo, [> `Msg of string]) result`.

Looks like the same to me and I'm not fond of delaying PR for another refactor..

NathanReb:
> Errors are "weird github url" and "dev-repo without host", so they are roughly the same kind of error as Unknown, I don't see why they should cause a hard failure.

Have you ever encountered any of these? I'd be curious to know what kind of dev-repo URL triggers this as I feel like the proper fix would be to just improve our parsing.
I see your point about the hard failure and I guess as long they are being logged it's okay to keep going on with these.

I just realized that you're filtering those packages during the `duniverse opam` step but I believe it would be better to do it during the `duniverse lock` step so that all the opam dependencies are defined in the `opam.sxp` and only the one that can be cloned into the `duniverse` are in the `dune.sxp`. This will also better cope with #23.

Could you please move both the filtering and the logging there? Then we can merge!

TheLortex:
> Have you ever encountered any of these? I'd be curious to know what kind of dev-repo URL triggers this as I feel like the proper fix would be to just improve our parsing.

I've never encountered any of these !

> Could you please move both the filtering and the logging there? Then we can merge!

Moved in the last rebase :)

avsm:
go for it, i'll review the whole tool later. I trust your judgement :-)

PR #17, opened 2019-04-23
Setup formatting of dune and OCaml files

NathanReb:
This PR adds dune formatting of OCaml and dune files. I formatted the whole repo and added a check to the CI script.

I used the `conventional` preset profile for `ocamlformat` and added a custom `margin=100` instead of the default `margin=80`. I'm not necessarily a huge fan of the code formatting but I think the advantages of auto-formatting greatly overcomes the style considerations.

Let me know what you think!


avsm:
wfm!

Julow:
This was merged fast !

The original code was closer to the `default` profile than `conventional`. But using more conventional formatting is good too.

PR #18, opened 2019-04-26
Improve CI

NathanReb:
This a WIP so please don't merge.

I'm trying to improve the CI to benefit more from docker cache. I'd like to take the occasion to experiment here to improve the ocaml CI scripts in general.

There are some more things I'd like to do before unWIPing this:
- [x] Make formatting check a separate build as @Julow smartly suggested.
- [ ] Make the script more generic so it can be reused elsewhere, ie pass build args through `PACKAGES`, `OCAML_VERSION` and so on to the `docker build`.
- [ ] Use more generic default build and test commands and expose a way to customize them and even run an `extra` step to fit everyone's need.


.dockerfiles/opam
```diff
@@ -0,0 +1,18 @@
...
+WORKDIR /home/opam/src
+USER opam
+COPY *.opam .
+RUN opam pin add --no-action duniverse .
+RUN opam update
+RUN opam depext duniverse ocamlformat=0.9
+RUN opam install ocamlformat=0.9
+RUN opam install --deps-only -t duniverse
+COPY . .
+RUN sudo chown -R opam /home/opam/src
```
Julow: `COPY` has a `--chown` option ([doc](https://docs.docker.com/engine/reference/builder/#copy)) so you can write this as:
```
COPY --chown=opam:opam . .
```

NathanReb: Uuuhh, neat!

.dockerfiles/opam
```diff
@@ -0,0 +1,18 @@
+ARG DISTRO
+FROM ocaml/opam2:${DISTRO}
+
+WORKDIR /home/opam/src
+USER opam
+COPY *.opam .
+RUN opam pin add --no-action duniverse .
+RUN opam update
+RUN opam depext duniverse ocamlformat=0.9
+RUN opam install ocamlformat=0.9
```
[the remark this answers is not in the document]
NathanReb: good catch thanks!

.dockerfiles/opam
NathanReb: Hmm it seems it's not working for some reason. Maybe it's not setting the ownership recursively :/

NathanReb: I'm a little surprised as it works perfectly fine when I'm running both `docker build` on my machine.

NathanReb: Also `alpine` seems unhappy about whatever docker translates the `COPY --chown=opam` to so I'll revert those changes, too bad.

NathanReb:
It seems like travis won't actually let us benefit from the docker cache, which is a damn shame if that's indeed the case.

I found an article mentioning that it's worth pulling your images at the begining of the build and pushing them at the end  to a docker hub or a an S3 bucket to effictively use the cache with travis.

I believe we're looking into moving away from travis so it might be a good time to experiment with other providers.

avsm:
There's actually already Docker support for Travis builtin to the ocaml-ci-scripts (see cohttp for an example).  However, it would be good to slowly move away from Travis to avoid being locked in.

One alternative in https://github.com/avsm/ocaml-yaml is the Azure CI setup, which tests this whole lot on macOS, Windows and Linux.  However, rather than copying those scripts, I have another idea: we could just add support to Dune itself to generate the CI boilerplate after https://github.com/ocaml/dune/pull/2091 is merged :-)

NathanReb:
I use `.travis-docker.sh` on a few projects of mine but I'm not a huge fan as I don't think it takes advantage of docker cache by doing the two step copy I.e.:
- copy the `.opam` files first and install dependencies
- copy the rest of the repo and build
Which allow you to use the cache instead of reinstalling everything everytime, which, even though it isn't really natively compatible with travis, is at least convenient to try the builds locally.

I agree it would be nice to let dune generate our CI scripts, I think I was more trying to experiment around what is it exactly that it should generate.

Also I'm curious to know what you think about the separate `ocamlformat` build which is, I believe, something we should aim at for the future of OCaml's CI.

NathanReb:
I'm unWIPing this as I think it's already an improvement over the current script. I'll try to contribute the separate formatting check to `ocaml-ci-scripts` as well.

To do that it would be nice to be able to host the `ocamlformat:VERSION` images on the `ocaml` docker hub instead of mine!

.docker.sh
```diff
@@ -1,30 +1,8 @@
...
- ./build-examples.sh)
-echo "Checking code formatting"
-make format
-if ! git diff --exit-code ; then
-  echo "Incorrect formatting, please run 'make format' before comitting changes"
-  exit 1
+if [ -z "$OCAMLFORMAT" ]; then
+  docker build --build-arg DISTRO=$DISTRO -f .dockerfiles/opam .
+else
+  docker build --build-arg OCAMLFORMAT=$OCAMLFORMAT -f .dockerfiles/ocamlformat .
```
Julow: I'm not fan of this script. It is here for having a clean `matrix` in `.travis.yaml` but it feels a bit magical to me.

What do you think of something like this instead:

```yaml
script: docker build --build-arg "$BUILD_ARG" -f "$DOCKERFILE" .
matrix:
  - BUILD_ARG=DISTRO=distro-1 DOCKERFILE=.dockerfiles/opam
  ...
  - BUILD_ARG=OCAMLFORMAT=0.9 DOCKERFILE=.dockerfiles/ocamlformat
```

NathanReb: The build arg part is easy to mess up so I'm not sure this would be an improvement. Is there a way we can improve the script so that it would feel a bit clearer what's happening?

NathanReb: Also `--build-arg` only takes one variable, you have to pass it repeatedly to set several variables so that wouldn't be extensible at all.

Julow: I see. Repeating script commands would be even worse. We could try using the "stage" feature from travis but I don't known how it works.

NathanReb: I'm not sure the stage would be much better but it could indeed work. What I like about the current design is you don't have to make an ordering choice between quality and regular builds, which you'd need to do with stages.

NathanReb:
Ok let's merge this as is and we improve it over time!

PR #20, opened 2019-05-10
Refactor Opam_cmd

NathanReb:
Small refactorings ahead of the merge of the opam and lock commands and files to keep the PRs size reasonable!

### DIFF
--- a/lib/dune_cmd.ml
+++ b/lib/dune_cmd.ml
@@ -107,7 +107,7 @@ let gen_dune_lock repo () =
   let ifile = Fpath.(repo // Config.opam_lockfile) in
   let ofile = Fpath.(repo // Config.duniverse_lockfile) in
   Opam.load ifile >>= fun opam ->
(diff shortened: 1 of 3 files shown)


lib/exec.ml
```diff
@@ -114,23 +114,23 @@ let git_add_all_and_commit ~repo ~message () =
 
 let git_merge ?(args = Cmd.empty) ~from ~repo () = run_git ~repo Cmd.(v "merge" %% args % from)
 
-let opam_cmd ~root sub_cmd =
+let opam_cmd ~opam_root sub_cmd =
```
Julow: Why this rename ? All these functions have `opam` in their name so `~root` is better IMO.

NathanReb: I wanted to disambiguate `root` and `roots` but indeed given the `opam` option is called `root` it makes sense to keep that way.

I reverted that and improved the naming for `roots` which is now `root_packages`. I also consistently renamed `pkg` to `package` in labelled arguments and record field and it indeed looks better!

PR #22, opened 2019-05-13
Mark the package as virtual when the url field is not understood

TheLortex:
Right now `ocamlfind` is classified as `Unknown` which leads to hard failure while executing `duniverse opam`. This is because `dev-repo` and url fields are filled but duniverse cannot do anything with them.
Instead of failing, I simply suggest to mark the package as `Virtual`, like it's done for other packages that are duniverse-incompatible.

This solves the problem I have introduced in #14.

### DIFF
--- a/lib/opam_cmd.ml
+++ b/lib/opam_cmd.ml
@@ -122,7 +122,13 @@ let classify_package ~package ~dev_repo ~archive ~pins () =
             | true ->
                 let base_repo = String.cuts ~empty:false ~sep:"#" archive |> List.hd in
                 (`Git base_repo, tag)
-            | false -> (`Unknown host, tag) )
+            | false ->
+                Logs.debug (fun l ->
+                    l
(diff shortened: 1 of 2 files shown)


lib/opam_cmd.ml
```diff
@@ -122,7 +122,10 @@ let classify_package ~package ~dev_repo ~archive ~pins () =
             | true ->
                 let base_repo = String.cuts ~empty:false ~sep:"#" archive |> List.hd in
                 (`Git base_repo, tag)
-            | false -> (`Unknown host, tag) )
```
Julow: You removed the only use of `Unknown`, you can remove it in `package_is_valid` below and in `types.ml`

TheLortex: Removed these entries !

TheLortex:
That's what I do in https://github.com/avsm/duniverse/pull/14

NathanReb:
This can be closed as we went for #14 as a solution instead!

PR #25, opened 2019-05-14
Merge opam and lock commands and opam.sxp and dune.sxp files

NathanReb:
This is just a first step to simplify the workflow. The new `init` command basically executes the old `opam` and `lock` sequentially to build a `duniverse.sxp` which contains the union of the fields from the old files.

Nothing smart for now but it opens up a path for improvement, especially about what goes into the `duniverse.sxp` and how we build it.

I have some plans for next steps but I wanted to keep this PR minimal.


NathanReb:
Just rebased!

cli/init.ml
```diff
@@ -0,0 +1,109 @@
...
+  let man = [] in
+  Term.info "init" ~doc ~exits ~man
+
+let term =
+  let open Term in
+  term_result
+    ( const run $ Common.Arg.repo $ Common.Arg.branch $ explicit_root_packages $ excludes $ pins
+    $ remotes $ Common.Arg.setup_logs () )
+
+let cmd = (term, info)
```
Julow: Why a new library ?
Half of `Cmdliner` code is still in `bin/duniverse.ml`, which contains only that plus `let () = Term.exit ...`
I'd personally prefer if `Cmdliner` code was in `bin/` and logic (eg. `Init.run`) in `lib/`.
This library is meant to be opened so I think the module names should be longer, eg. `cli_common.ml` and `init_cmd.ml`.

lib/git_cmd.ml
```diff
@@ -1,70 +0,0 @@
...
-let merge repo branch () =
-  let vendor_branch = Config.vendor_branch branch in
-  Logs.app (fun l ->
-      l "%aGit checking out %a branch." pp_header header Fmt.(styled `Cyan string) branch );
-  Exec.git_checkout ~repo branch >>= fun () ->
-  Logs.app (fun l ->
-      l "%aGit merging from %a branch" pp_header header Fmt.(styled `Cyan string) vendor_branch );
-  Exec.git_merge ~from:vendor_branch ~args:Cmd.(v "--squash") ~repo () >>= fun () ->
-  let message = Fmt.strf "update vendor libraries from branch %s" vendor_branch in
-  Exec.git_add_all_and_commit ~repo ~message ()
```
Julow: Unrelated change but was dead code.

cli/init.ml
```diff
@@ -0,0 +1,109 @@
...
+  let open Rresult.R.Infix in
+  Common.Logs.app "Calculating Duniverse on the %a branch" Styled_pp.branch branch;
+  compute_opam ~repo ~branch ~explicit_root_packages ~excludes ~pins ~remotes >>= fun opam ->
+  let { Opam.root_packages; excludes; pins; remotes; branch; packages } = opam in
+  Common.Logs.app "Calculating Git repositories to vendor";
+  compute_repos ~packages >>= fun repos ->
+  let duniverse = { Duniverse.root_packages; excludes; pins; remotes; branch; packages; repos } in
+  let file = Fpath.(repo // Config.duniverse_file) in
+  Duniverse.save ~file duniverse >>= fun () ->
+  Common.Logs.app "Wrote duniverse file with %a entries to %a."
```
Julow: Colors are not working, maybe because messages are converted to string by `Common.Logs.app`.

cli/init.ml
```diff
@@ -0,0 +1,109 @@
...
+    Fmt.(styled `Green int)
+    (List.length repos)
+    Fmt.(styled `Cyan Fpath.pp)
+    file;
```
Julow: Path is printed as `./duniverse.sxp`. You should use `Fpath.normalize` to clean the path before printing.

cli/init.ml
NathanReb: Uh good catch!

cli/init.ml
NathanReb: Nice tip! I'll do that!

NathanReb:
As a follow up to our offline discussion I'll split some of the refactorings here to separate PRs!

NathanReb:
@avsm please don't merge right away there's a couple points that need fixing!

NathanReb:
Should be good now!

Next step is to refactor the `Duniverse` module so that instead of the two `packages` and `repos` field we have a single one that describe the whole universe with some kind of variant which would look kind of like:
```ocaml
type opam_package =
  { name : string; version : string }

type dep =
  | Opam of opam_package
  | Repo of
    { opam_packages: opam_package list
    ; upstream : upstream
    ; fork : upstream option
    ...
    }
```

with the mid/long term objective of getting rid of the `Opam` variant once we can deal with any dependency.

avsm:
Nice one! Merging and can iterate in trunk

PR #27, opened 2019-05-20
Extract Duniverse_cli

NathanReb:
This PR organizes the CLI tool a bit better. In particular I extracted a `Duniverse_cli` library that contains one module per command and a `Common` module for shared options or arguments.

For now I just moved some code around, removed dead code, renamed a few variables and removed the extra `unit` argument of the library versions as it was only there to enable passing of common arguments such as `--verbose`.

The long term idea is to group together the declarative `Cmdliner` code that is specific to a command and, once we have better `Duniverse_lib` functions to have the actual command logic as a single `run` function that just defines the sequence of operations it runs (e.g. read config file, extract some field, apply a library function to filer it, write some other file).


avsm:
Thanks!

PR #29, opened 2019-05-27
Download cache

TheLortex:
An attempt to solve https://github.com/avsm/duniverse/issues/15

What it does:
* Create a `.duniverse` in `$HOME` containing a git repository in a `cache` subdirectory.
* When pulling a package from a given `upstream` with a given `tag`, it checks if the `upstream-tag` branch already exists. If it does, a simple directory to directory git clone is done from it. If not it add the `upstream` to the list of upstreams and fetchs the requested `tag`.

This should allow for the same package to have multiple upstreams/tags. 

However the cache assumes that a `tag` doesn't move, so packages that rely on a branch (such as master) are problematic for now.


avsm:
This is good, but the branch thing is a definite problem since we pin quite often to a branch.  I have a slightly modified idea.  How about instead of ever having tags in the duniverse.sxp, we only ever have sha256 commit hashes (as esy does).  Those are straightforward to cache using your scheme, and are immutable.

We'll need to modify the opam phase to resolve URLs to a concrete sha256 tip though, before this can be merged.

lib/cache.ml
```diff
@@ -0,0 +1,59 @@
...
+                output_dir );
+          Ok ()
+      | false -> (
+          Logs.app (fun l ->
+              l "%aPulling sources for %a." pp_header header Fmt.(styled `Cyan Fpath.pp) output_dir
+          );
+          git_remote_exists_or_create ~remotes ~remote >>= function
+          | () -> Exec.git_fetch_to ~remote_name ~tag ~branch ) )
+    ()
+  |> Rresult.R.join
```
avsm:
```suggestion
  |> R.join
```

lib/cache.ml
```diff
@@ -0,0 +1,59 @@
...
+
+let git_get ~output_dir ~remote ~tag () =
+  let home = Unix.getenv "HOME" in
```
avsm: Bos.OS.Env.req_var will output a good error message rather than Unix.getenv (which will raise Not_found)

lib/cache.ml
```diff
@@ -0,0 +1,59 @@
...
+  let cache_location = Fpath.(v home / ".duniverse" / "cache") in
```
avsm: Probably worth plumbing through `DUNIVERSE_CACHE` env var to let the user set it easily.  In CI, we can map it to a volume mount in a docker container then to be shared easily across CI runs.

NathanReb:
> This is good, but the branch thing is a definite problem since we pin quite often to a branch. I have a slightly modified idea. How about instead of ever having tags in the duniverse.sxp, we only ever have sha256 commit hashes (as esy does). Those are straightforward to cache using your scheme, and are immutable.
>
> We'll need to modify the opam phase to resolve URLs to a concrete sha256 tip though, before this can be merged.

Yep that's what we came up with @TheLortex as well a few days ago when discussing this feature. I have to get some stuff done on the `dune 1.10` X `dune-release` front first but I'm planning on a refactoring of the `duniverse` module to better represent the duniverse in order to make it easier to pin a tracked branch to a commit ref.

The idea would be to let you track a branch of an upstream repo at the time of `duniverse init` or the future `duniverse install`. This would be tracked in the `duniverse.sxp` along with the commit ref for this branch's `HEAD` at the time of execution. You could then later "upgrade" such repos of your duniverse by running `duniverse update` or something similar which would update the commit refs for those branches in the `duniverse.sxp` and pull in the latest versions in your `duniverse/`.

avsm:
That scheme sounds perfect @NathanReb.  Tracking the 'name' of the source code origin (the url/branch/tag name) and having the 'resolved' address (the hash) as separate concerns would work well.  In the future, we can map changes to local code back by following the names of the relevant directories back.

lib/cache.ml
samoht: Also it's worth using XDG conventions to avoid cluttering HOME with dot files/dirs. For instance using: 
```
$XDG_CACHE_HOME defines the base directory relative to which user specific non-essential data files should be stored. If $XDG_CACHE_HOME is either not set or empty, a default equal to $HOME/.cache should be used.
```
(from https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

emillon: FYI there's some code in dune that implements this.

TheLortex:
@NathanReb 
> I'm a bit confused by this approach and I find the code a bit hard to follow.
> 
> I'm especially surprised by the fact that the `cache` directory a git repo and we use only branches to track repos, upstreams and refs. Could you explain the rationale behind this choice a bit? I feel like my git-fu isn't strong enough and I'm lagging behind a little!
> 
> I added a few minor comments but otherwise it looks good!

The scheme is the following, given a `remote_url` and a `tag`.
* the pair `remote_url,tag` is translated to a unique branch name
* then a simple lookup checks if the branch already exists in the cache (using direct file lookup in order to spare another `git` call):
  - if it doesn't exist: check if the remote already exists:
    * if it doesn't exist: create the remote
    * shallow fetch the target `tag` and create the branch from it
  - clone the branch to the desired directory

Is there something that you find problematic in that ?

lib/exec.ml
```diff
@@ -177,3 +174,27 @@ let run_opam_install ~yes packages =
...
+let git_init output_dir = run_and_log Cmd.(v "git" % "init" % p output_dir)
+
+let git_clone ~branch ~remote ~output_dir =
+  run_and_log
+    Cmd.(v "git" % "clone" % "--depth" % "1" % "--branch" % branch % remote % p output_dir)
+
+let git_rename_branch_to ~branch = run_and_log Cmd.(v "git" % "branch" % "-m" % branch)
+
+let git_remotes dir =
+  OS.Dir.with_current dir
```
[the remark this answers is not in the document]
TheLortex: I added the `-C` option but haven't extracted a `git` function yet.

lib/cache.ml
TheLortex: So now it checks for a `DUNIVERSE_CACHE`, then for a `XDG_CACHE_HOME` and finally uses `HOME` to find a cache location !

lib/cache.ml
```diff
@@ -0,0 +1,68 @@
...
+              l "%aPulling sources for %a." pp_header () Fmt.(styled `Cyan Fpath.pp) output_dir );
+          git_remote_exists_or_create ~remotes ~remote >>= function
+          | () -> Exec.git_fetch_to ~remote_name ~tag ~branch ) )
+    ()
+  |> R.join
+
+let cache_dir =
+  let ( / ) = Filename.concat in
+  let default_cache_dir () =
+    let os_cache_dir = if Sys.win32 then "Local Settings" / "Cache" else ".cache" in
```
Julow: You are handling the windows case here but still using the `HOME` variable, which will not be set. I think you should use the variable `UserProfile` (to have the `Local Settings` directory) or `AppData`.

lib/cache.ml
```diff
@@ -0,0 +1,68 @@
...
+  let remote_name = valid_remote_name remote in
+  match String.Set.mem remote_name remotes with
+  | true -> Ok ()
+  | false -> Exec.git_remote_add ~remote_url:remote ~remote_name
+
+let git_branch_exists_or_create ~output_dir ~remotes ~branch ~remote ~tag cache_dir =
+  let remote_name = valid_remote_name remote in
+  OS.Dir.with_current cache_dir
+    (fun () ->
+      OS.Path.exists Fpath.(v ".git" / "refs" / "heads" / branch) >>= function
```
samoht: this will not work with `packed-refs` -- e.g. after you run a `git gc`. Usually it's really unsafe to look in `.git/` directly as the Git tool will often optimise things behind your back.

lib/cache.ml
```diff
@@ -0,0 +1,68 @@
...
+    let os_cache_dir = if Sys.win32 then "Local Settings" / "Cache" else ".cache" in
+    Bos.OS.Env.req_var "HOME" >>| fun home -> home / os_cache_dir
+  in
+  let xdg_cache_dir () =
+    match Bos.OS.Env.var "XDG_CACHE_HOME" with Some x -> Ok x | None -> default_cache_dir ()
+  in
+  (match Bos.OS.Env.var "DUNIVERSE_CACHE" with Some x -> Ok x | None -> xdg_cache_dir ())
+  >>= fun cache_dir -> Fpath.of_string (cache_dir / "duniverse")
+
+let git_get ~output_dir ~remote ~tag () =
```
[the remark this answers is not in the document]
TheLortex: Sure!

lib/cache.ml
TheLortex: To solve this I'm using the same strategy as dune:
https://github.com/ocaml/dune/blob/master/src/xdg/xdg.ml

TheLortex:
Rebased and updated according to reviews!

lib/cache.ml
```diff
@@ -0,0 +1,76 @@
...
+  OS.Dir.delete ~recurse:true output_dir >>= fun () ->
+  Exec.git_clone ~branch:cache_branch_name ~remote:(Fpath.to_string cache_location) ~output_dir
+
+(* Setup usable remotes in the cloned directory *)
+let setup_remote ~output_dir ~remote ~tag () =
+  Exec.git_rename_branch_to ~repo:output_dir ~branch:tag >>= fun () ->
+  Exec.git_remote_add ~repo:output_dir ~remote_name:"upstream" ~remote_url:remote >>= fun () ->
+  Exec.git_remote_remove ~repo:output_dir ~remote_name:"origin"
+
+let git_get ~output_dir ~remote ~tag () =
```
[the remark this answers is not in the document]
TheLortex: Yes: the goal is to be able to make local modifications and push them using git

cli/pull.ml
```diff
@@ -68,7 +68,7 @@ let pull_source_dependencies ~duniverse_dir src_deps =
     (fun { Duniverse.Deps.Source.dir; upstream; ref; _ } ->
       let output_dir = Fpath.(duniverse_dir / dir) in
       Common.Logs.app (fun l -> l "Pulling sources for %a." Styled_pp.path output_dir);
-      Exec.git_archive ~output_dir ~remote:upstream ~tag:ref () )
+      Cache.git_get ~output_dir ~remote:upstream ~tag:ref ())
```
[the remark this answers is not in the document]
TheLortex: Now it prints `[CACHED]` when the content has been fetched from cache.

lib/cache.ml
```diff
@@ -0,0 +1,76 @@
...
+  in
+  Fpath.of_string (cache_dir () / "duniverse")
+
+(* Check if the cache repository is initialized and if not, initialize it *)
+let check_duniverse_cache_directory cache_location =
+  OS.Dir.exists cache_location >>= function
+  | true -> Ok ()
+  | false ->
+      OS.Dir.create cache_location >>= fun _ ->
+      Exec.git_init ~repo:cache_location
```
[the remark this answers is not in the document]
TheLortex: I haven't been able to find a way to fetch an annotated tag directly into a branch on bare repos, so for now a worktree is actually needed.

lib/cache.ml
```diff
@@ -0,0 +1,73 @@
...
+open Astring
+
+let valid_remote_name = String.map (function '/' | '@' | ':' -> '-' | x -> x)
+
+let git_remote_exists_or_create ~repo ~remote =
+  let remote_name = valid_remote_name remote in
+  Exec.git_remotes ~repo >>= fun remotes ->
+  match List.exists (String.equal remote_name) remotes with
+  | true -> Ok ()
+  | false -> Exec.git_remote_add ~repo ~remote_url:remote ~remote_name
```
Julow: Haha this could be an `if`.

lib/cache.ml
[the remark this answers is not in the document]
Julow: This is unrelated to the download cache. I suggest keeping the same behavior and changing it when everything is ready for it.

lib/cache.ml
```diff
@@ -0,0 +1,73 @@
...
+  in
+  let xdg_cache_dir () =
+    match Bos.OS.Env.var "XDG_CACHE_HOME" with Some x -> x | None -> default_cache_dir ()
+  in
+  let cache_dir () =
+    match Bos.OS.Env.var "DUNIVERSE_CACHE" with Some x -> x | None -> xdg_cache_dir ()
+  in
+  Fpath.of_string (cache_dir () / "duniverse")
+
+(* Check if the cache repository is initialized and if not, initialize it *)
```
Julow: This and the following are not doc comments (`(**`).

lib/cache.mli
```diff
@@ -0,0 +1,7 @@
+val git_get :
+  output_dir:Fpath.t -> remote:string -> tag:string -> unit -> (bool, Rresult.R.msg) result
+(** [git_get ~output_dir ~remote ~tag ()]] fetch a git project from the given remote at the given tag and place it in the output_dir.
+    It uses a cache mechanism to locally fetch the (remote, tag) pair when it's already available. The cache assumes that the given tag
+    won't be updated on the remote. output_dir dans it's content will be removed if it exists.
```
Julow: typo `output_dir dans`, `and` ?

lib/cache.ml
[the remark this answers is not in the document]
Julow: I did not manage to do it as a single `git fetch` and got some weird errors (eg `error: cannot update ref '...': trying to write non-commit object ...`).
But the `FETCH_HEAD` trick is also available in bare repo: `git -C $cache_location --bare branch $branch FETCH_HEAD`.

lib/cache.ml
TheLortex: oop!

lib/cache.ml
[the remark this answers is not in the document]
TheLortex: Right indeed it's possible to have everything under a bare directory. I'll work on that !

lib/cache.ml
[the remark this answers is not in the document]
TheLortex: I'll separate the PRs !

TheLortex:
As branches are now resolved into commit hashes, the cache mechanism has been updated accordingly. Basically for each target commit a `commit-<hash>` branch is created that can directly be cloned into the target directory.

TheLortex:
Rebased on `master` ! This means the cache is now parallel, for even more speed !!
What changed is that instead of using a single git repo for every remote, there's now one repo per remote.

cli/pull.ml
```diff
@@ -64,31 +65,19 @@ let mark_duniverse_content_as_vendored ~duniverse_dir =
...
-  Exec.git_rev_parse ~repo:output_dir ~ref:"HEAD" () >>= fun head_commit ->
-  checkout_if_needed ~head_commit ~output_dir ~ref ~commit ()
+  Cache.clone_to ~output_dir ~remote:upstream ~ref ~commit ()
   |> Rresult.R.reword_error (fun (`Msg _) -> `Commit_is_gone dir)
-  >>= fun () ->
+  >>= fun cached ->
+  if cached then
+    Common.Logs.app (fun l ->
+        l "Pulled sources for %a. %a" Styled_pp.path output_dir Styled_pp.cached () )
+  else Common.Logs.app (fun l -> l "Pulled sources for %a." Styled_pp.path output_dir);
```
Julow: The `if` could be moved to `Styled_pp` that would be a `bool Fmt.t`:
```ocaml
let cached pp cached = if cached then string pp " [CACHED]" else ()
```

lib/cache.ml
```diff
@@ -0,0 +1,88 @@
...
+let commit_branch_name ~commit = "commit-" ^ commit
+
+let cache_branch_name ~ref = "branch-" ^ ref
+
+let home () =
+  try Sys.getenv "HOME"
+  with Not_found -> (
+    try (Unix.getpwuid (Unix.getuid ())).Unix.pw_dir
+    with Unix.Unix_error _ | Not_found ->
+      if Sys.win32 then try Sys.getenv "AppData" with Not_found -> "" else "" )
```
Julow: The `if Sys,win32` should be done before trying `Unix.getpwuid`.

lib/exec.ml
```diff
@@ -126,6 +126,46 @@ let git_resolve ~remote ~ref =
       Rresult.R.error_msgf "Multiple refs share the name %a on the remote %s" Git.Ref.pp ref remote
   | Error (`Msg _) as err -> err
 
+let git_resolve_local ~repo ~ref =
+  run_and_log_l Cmd.(v "git" % "-C" % p repo % "show-ref" % ref) >>= fun output ->
```
Julow: Is this command is expected to fail ?
If yes, errors should not be logged as ERROR (`~ignore_error:true`).
I'm getting 2 screens of:

```txt
duniverse: [ERROR] git -C
                     /home/jules/.cache/duniverse/https---github.com-janestreet-sexplib0.git
                     show-ref branch-v0.12.0 failed. Output was:
```

lib/cache.ml
NathanReb: I'll fix that!

cli/update.ml
```diff
@@ -15,7 +15,9 @@ let update ~total ~updated src_dep =
   let open Result.O in
   let ref : Duniverse.resolved = src_dep.Duniverse.Deps.Source.ref in
   incr total;
-  Exec.git_resolve ~remote:src_dep.upstream ~ref:ref.t >>= fun new_ref ->
+  Cache.update ~remote:src_dep.upstream ~ref:ref.t ()
```
NathanReb: I don't think that does the right thing either, this seems to update the cache, which is not what we want here. We want to update the commit hash corresponding to the tracked ref in the dune-get.

Julow:
Not using the cache for ref resolution (eg. using the cache only when pulling) definitely makes sense as a first step.

cli/init.ml
```diff
@@ -31,7 +31,7 @@ let compute_deps ~opam_entries =
   Duniverse.Deps.from_opam_entries ~get_default_branch opam_entries
 
 let resolve_ref deps =
-  let resolve_ref ~upstream ~ref = Exec.git_resolve ~remote:upstream ~ref in
+  let resolve_ref ~upstream ~ref = Cache.resolve ~remote:upstream ~ref () in
```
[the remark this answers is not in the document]
Julow: Caching resolution may not even be necessary since `init` should not be called often and resolution should be fast enough.

NathanReb:
I rebased this and pushed new commits with some of the changes we suggested!

We had a look at this with @Julow and came to the conclusion that part of the cache directory finding logic is a bit too intrusive and we suggest to do the following:
1. Set the cache dir to the path pointed at by the `$DUNIVERSE_CACHE` env var
2. If it's not set, set it to `$XDG_CACHE_HOME`
3. If it's not set, set it to `$HOME\Local Settings\Cache` on windows or `$HOME/.cache` on Unix
4. If `$HOME` is not set, and we're on windows, set it to `$AppData\Local Settings\Cache`
5. If none of the above are set, we disable the cache and warn the user with a helpful error msg

@avsm how do you feel about that? I feel like it's a good default that should work for most people while not offending anyone.

emillon:
For the cache location, you can reuse [dune's xdg library](https://github.com/ocaml/dune/tree/master/src/xdg) which seems to do something similar to what you're describing.

Julow:
@emillon Dune is using `Unix.getpwuid` and we found that to be a bit intrusive. In the case `HOME` is not set, we instead disable the cache.
What do you think?

emillon:
I wouldn't call that intrusive, but sure it works to disable the cache when HOME is not set.

My point was more that you can vendor `xdg` into duniverse and not have to worry about these details :)

NathanReb:
I played with the new cache a bit, fixed a bug, added a `--no-cache` option to disable the cache (that can be helpful to debug some issues) so I think it's now ready for a final round of review.

Let's get this merged!

cli/pull.ml
```diff
@@ -136,7 +125,11 @@ let run yes repo () =
       let duniverse_dir = Fpath.(repo // Config.vendor_dir) in
       Bos.OS.Dir.create duniverse_dir >>= fun _created ->
       mark_duniverse_content_as_vendored ~duniverse_dir >>= fun () ->
-      pull_source_dependencies ~duniverse_dir duniverse
+      get_cache ~no_cache >>= fun cache -> pull_source_dependencies ~duniverse_dir ~cache duniverse
+
+let no_cache =
+  let doc = "Run without using the duniverse global cache" in
```
avsm: Before merging, could the help be improved so that the actual behaviour is described here? In particular, the environment variables that are looked up so that user's know what's going on and how to modify it.

NathanReb: Yeah really good point! I'll improve the man page!

NathanReb: I added some documentation about that, let me know what you think!

cli/pull.ml
```diff
@@ -136,24 +125,61 @@ let run (`Yes yes) (`Repo repo) () =
...
+  in
   let man =
     [ `S Manpage.s_description;
       `P
         "This command reads the Git metadata calculated with $(i,duniverse lock) and fetches them \
          from their respective Git remotes and stores them in the $(b,duniverse/) directory in \
-         the repository."
+         the repository.";
+      `P
+        "This command uses a global dunivere cache to avoid repeated downloads. \
```
avsm:
```suggestion
        "This command uses a global duniverse cache to avoid repeated downloads. \
```

lib/cloner.ml
```diff
@@ -0,0 +1,138 @@
...
+  let clone_to ~output_dir ~remote ~ref ~commit cache_dir =
+    let open Result.O in
+    get_cache_directory ~remote cache_dir >>= fun repo ->
+    check_package_cache_branch ~repo ~ref ~commit () >>= clone_from_cache ~output_dir ~repo
+end
+
+module Uncached = struct
+  let warn_about_head_commit ~ref ~commit () =
+    Logs.info (fun l ->
+        l "%a is not the HEAD commit for %a anymore" Styled_pp.commit commit Styled_pp.branch ref
```
avsm: Not a blocker for this PR, but having a warning about HEAD could be annoying if you are deliberately pinning a package. There's no way to disable this warning as far as I can tell.

NathanReb: This actually is an info log, which means it is disabled by default and you'll only see it if you run with the `-v` option. I believe it can be valuable information in a verbose output as it can help you figure out why you're not in sync with the tracked branch after a pull!

What do you think?

We actually changed it in #53 I believe, it used to be a warning before!

PR #30, opened 2019-06-02
Various tweaks: rename default config file to dune-get and help fixes

avsm:


Julow:
Why this change? I think the `duniverse/` folder name and the tool name should be consistent with it.

avsm:
I quite like the vendor directory remaining `duniverse/` as a quirky ocaml name, even when merged into dune :-)

The intention of changing the config file is purely to get it aligned with dune conventions (dune-project, dune, dune-get, ...).  I expect that we will eventually drop `dune-get` and directly modify `dune-project` in order to have metadata all in one place. See for example #31 where we need more fine-grained information from the overall dune-project files in the future.

avsm:
I'm also ok with leaving the prototype tool named `duniverse` for the moment @NathanReb.  I just wanted to get the artefacts on disk a bit more natural (i.e. no `.sxp` extensions) as I'd like duniverse to self-bootstrap.

PR #32, opened 2019-06-04
Improve tag_from_archive

NathanReb:
This PR add a couple test and minor robustness improvements to the parsing of archive URLs.

It used to crash on empty archive fields and although this should not happen, better be safe than sorry.


avsm:
lgtm, i've hit that empty archive crash myself!

PR #33, opened 2019-06-04
Add test for classify_package

NathanReb:
This function is quite important and it's pretty hard to understand what it does so I added test both to prevent breakage and characterize its behaviour.

It probably needs some fix from the look of it so we can discuss this here before merging.

My initial plan was to drop the `Github` variant and just detect git based dev-repo as that should be all `duniverse` need to know since the tag/branch are guessed from the archive URL. When I looked into this function I figured it might be a bad idea to blindly remove the `Github` case in its current state but I believe we can refine this function so this can be achieved.


avsm:
I think removing Github should be fine.  The only use of it now looks very similar to Git:

```
  match opam.Opam.dev_repo with
  | `Github (user, repo) -> (
      let upstream = Fmt.strf "https://github.com/%s/%s.git" user repo in
      match opam.Opam.tag with
      | None ->
          Exec.git_default_branch ~remote:upstream () >>= fun ref -> Ok { Dune.dir; upstream; ref }
      | Some ref -> Ok { Dune.dir; upstream; ref } )
  | `Git upstream -> (
    match opam.Opam.tag with
    | None ->
        Exec.git_default_branch ~remote:upstream () >>= fun ref -> Ok { Dune.dir; upstream; ref }
    | Some ref -> Ok { Dune.dir; upstream; ref } )
```

and most of the gory logic is in opam_cmd to guess the tag name from the archive, which doesnt use the Github variant.

NathanReb:
Yep that's why I thought about removing it but given what gets picked as the repo URL in the `Git` case I'd like to improve this function first!

test/test_opam_cmd.ml
```diff
@@ -40,3 +40,55 @@ let test_tag_from_archive =
...
+      ~dev_repo:"git+https://host.com/some-repo" ~archive:""
+      ~expected:(`Unknown "host.com", None)
+      ();
+    make_test ~name:"no host" ~package:{ name = "x"; version = None } ~dev_repo:"nohost"
+      ~archive:""
+      ~expected:(`Error "dev-repo without host", None)
+      ();
+    make_test ~name:"git" ~package:{ name = "x"; version = None }
+      ~dev_repo:"git+https://host.com/some-repo" ~archive:"gitpaf#pouf"
+      ~expected:(`Git "gitpaf", None)
```
NathanReb: See this test, I believe it uses the archive where it should use the `dev-repo` and that's a bug @avsm

lib/types.ml
```diff
@@ -33,6 +33,15 @@ module Opam = struct
     | `Error of string ]
   [@@deriving sexp]
 
+  let pp_repo = pp_sexp sexp_of_repo
+
+  let equal_repo r r' =
```
Julow: Why not using `[@@deriving eq]` ?

test/test_opam_cmd.ml
```diff
@@ -40,3 +40,55 @@ let test_tag_from_archive =
     make_test ~archive:"https://other.domain.com/some/path/file-v1.2.3.tbz"
       ~expected:(Some "v1.2.3") ()
   ]
+
+let test_classify_package =
+  let make_test ~name ~package ?(dev_repo = "dummy-dev-repo") ?archive ?(pins = []) ~expected () =
```
Julow: Why not taking `~name` and `?version` arguments instead of `~package` ?

test/test_opam_cmd.ml
```diff
@@ -40,3 +40,55 @@ let test_tag_from_archive =
...
+      ~dev_repo:"git+https://host.com/some-repo.git" ~archive:""
+      ~expected:(`Unknown "host.com", None)
+      ();
+    make_test ~name:"no host" ~package:{ name = "x"; version = None } ~dev_repo:"nohost"
+      ~archive:""
+      ~expected:(`Error "dev-repo without host", None)
+      ();
+    make_test ~name:"git" ~package:{ name = "x"; version = None }
+      ~dev_repo:"git+https://host.com/some-repo.git" ~archive:"gitpaf#pouf"
+      ~expected:(`Git "https://host.com/some-repo.git", None)
```
Julow: This test is not passing:
```
[failure] Error git: expecting
((Git https://host.com/some-repo.git), None), got
((Git gitpaf), None).
```

test/testable.ml
```diff
@@ -0,0 +1,5 @@
+module Opam = struct
```
Julow: Shouldn't this be defined in `test_opam_cmd.ml` ? I think `Opam_cmd` as a whole is not unit testable so this will probably never get shared between tests.

NathanReb:
I'm working on fixing the bug atm thus the WIP status of this PR. Sorry about that, I'll ping you once it's ready for review again!

test/test_opam_cmd.ml
NathanReb: This is just to stick as much as possible to the signature of the function understand and avoid adding logic to the test as much as possible. It makes it easier to guess what the test will do i.e. just pass the arguments along to the function.

I agree in this case it seems acceptable but I'm not sure we'd gain much.

lib/types.ml
NathanReb: I did that while fixing the bug!

test/test_opam_cmd.ml
NathanReb: Now it is!

test/test_opam_cmd.ml
NathanReb: I implemented a fix and updated the test cases, let me know if it does what it should or if I got it wrong.

test/testable.ml
NathanReb: The idea is that this type can be used elsewhere so I added a `Testable` module so different tests can reuse that but I agree it might be a little bit too early and since it's only used in a single place I'll move it back!

duniverse.opam
```diff
@@ -11,6 +11,7 @@ depends: [
   "ocaml" {>= "4.05.0"} 
   "dune" {build >= "1.2.0"}
   "ppx_sexp_conv"
+  "ppx_deriving"
```
emillon: Is the end game still merging this into `dune`? If that's the case I don't think it's a good idea to add this dependency.

NathanReb: Does `stdune` have some helpers for that? It's a bit of a hassle to write those by hand in the meantime.

NathanReb:
I pushed some new commits to fix the bug, here's a summary of what I did:
- I extracted a `Git` module for git related helper functions, I added a `is_git_uri` function there that checks whether an uri is a git URI (for now just checks the path part as a `.git` extension. I also added tests for this function
- I extracted an `Opam` module for opam related helper functions. I added a `Dev_repo` module there, with a type and a `from_string` function that parses a dev-repo opam field value to a vcs and a cleaned up URI. I also added tests for this function.
- `classify_package` now uses `Opam.Dev_repo.from_string` to distinguish git based dev-repos and uses the cleaned URI in the git case
- I removed the Unknown variant as it's not used anymore

The next step which is to get rid of the `Github` variant will further simplify this function.

duniverse.opam
emillon: Not really. There are some extra types like `Ordering.t`, but the comparisons are written by hand. It's true that it's a lot of boilerplate to add, but either you'll have to do it now or later. So it's really your call :)

lib/git.ml
```diff
@@ -0,0 +1,4 @@
+let is_git_uri uri =
```
Julow: Git uri are not terminated by `.git`. They can take a lot of forms and we should not try to parse them. I suggest to move this function closer to where it is used because it does not make sense on its own and is only used to make guess.

lib/opam_cmd.ml
```diff
@@ -101,24 +101,21 @@ let classify_package ~package ~dev_repo ~archive ~pins () =
...
+            match Uri.host dev_repo_uri with
+            | Some "github.com" -> (
+              match String.cuts ~empty:false ~sep:"/" (Uri.path dev_repo_uri) with
+              | [ user; repo ] ->
+                  let repo = strip_ext repo in
+                  (`Github (user, repo), tag)
+              | _ -> err "weird github url" )
+            | Some _host -> (`Git (Uri.to_string dev_repo_uri), tag)
+            | None -> err "dev-repo without host" )
+          | { vcs = _; _ } -> (`Error "dev-repo doesn't use git as a VCS", None) ) )
```
Julow: Did you mean `{ vcs = None; _ }`?

lib/types.ml
```diff
@@ -25,14 +25,19 @@ module Opam = struct
...
+  type repo = [ `Github of string * string | `Git of string | `Virtual | `Error of string ]
   [@@deriving sexp]
 
+  let equal_repo repo repo' =
+    match (repo, repo') with
+    | `Github (user, repo), `Github (user', repo') ->
+        String.equal user user' && String.equal repo repo'
+    | `Git s, `Git s' | `Error s, `Error s' -> String.equal s s'
+    | `Virtual, `Virtual -> true
+    | (`Github _ | `Git _ | `Unknown _ | `Virtual | `Error _), _ -> false
```
Julow: You removed the `Unknown` tag.

test/test_opam_cmd.ml
```diff
@@ -40,3 +46,55 @@ let test_tag_from_archive =
...
+      ~dev_repo:"git+https://github.com/user/repo/oh/my/what/is/this" ~archive:""
+      ~expected:(`Error "weird github url", None)
+      ();
+    make_test ~name:"no host" ~package:{ name = "x"; version = None } ~dev_repo:"nohost.git"
+      ~archive:""
+      ~expected:(`Error "dev-repo without host", None)
+      ();
+    make_test ~name:"git" ~package:{ name = "x"; version = None }
+      ~dev_repo:"git+https://host.com/some-repo.git" ~archive:"gitpaf#pouf"
+      ~expected:(`Git "https://host.com/some-repo.git", None)
```
Julow: Shouldn't the tag be `Some "pouf"` ?

test/test_opam_cmd.ml
```diff
@@ -40,3 +46,55 @@ let test_tag_from_archive =
...
+      ();
+    make_test ~name:"wrong vcs" ~package:{ name = "x"; version = None }
+      ~dev_repo:"hg+https://host.com/some-repo" ~archive:""
+      ~expected:(`Error "dev-repo doesn't use git as a VCS", None)
+      ()
+  ]
```
Julow: Can you add tests with no "git+" prefix with and without the ".git" extension?

lib/opam_cmd.ml
NathanReb: No I meant `{ vcs = _; _}` but I guess that could just be replaced by `_`

lib/git.ml
NathanReb: Maybe we could rename it `has_git_extension`, in which case it describes exactly what happens and I think it's then fairly harmless to leave it here.

lib/types.ml
NathanReb: Good catch. This showcases why we should use regular variants here, especially since we don't take advantage of their polymorphic nature anywhere in the code.

test/test_opam_cmd.ml
NathanReb: No because the URL doesn't look like a git one to `tag_from_archive` so it doesn't try to parse the `#<branch_name>` as so, which is expected behaviour.

That doesn't mean `tag_from_archive` is perfectly robust but at least in this case it got it right!

test/test_opam_cmd.ml
NathanReb: These cases are tested in `Test_opam.Dev_repo` already, no point in duplicating them here.

lib/opam_cmd.ml
NathanReb: I made the case explicit: `{vcs = (None | Some (Other _)); _}`

lib/git.ml
Julow: I don't like keeping it here because it adds a new module, new test files and the presence of the ".git" extension does not mean anything to git.

NathanReb: Why would adding new modules or tests be a liablility?

It might not mean anything to git but it still is widely used, even https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols puts one in every single examples.

My take on this is having a module for git related helper functions make sense no matter what. This is just a step in a refactoring and with time more functions will most certainly end up here but even they don't that doesn't mean we most not create modules to separate concerns. This makes it easier when working on the `duniverse` repo to look for them if they seat in the right place.

NathanReb: To sum up our offline discussion: we'll move this function to a `Uri_utils` as a `Git` module would be better suited for functions that apply in a git context and this is used to determine whether we are in a git context or not.

NathanReb:
@Julow fixed that last point, let me know if it looks good to you and let's merge!

PR #34, opened 2019-06-05
Draft a new spec for duniverse

NathanReb:
This is a WIP. I'm opening the PR so we can discuss those the spec!

cc @samoht @avsm 

Not much as happened since it was posted on slack but I'll keep you updated whenever I make significant changes.

### DIFF
--- a/doc/spec.md
+++ b/doc/spec.md
@@ -1,222 +1,174 @@
-This document the `duniverse` tool aims and next steps.
+# Duniverse
 
-Overview
---------
+## High level spec
 
-The role of the duniverse tool is to setup a development environment for
-the dune build system, or more precisely a "workspace". Its
-distinguishing characteristic is that it can setup workspaces composed
-of multiple projects, bringing hassle free multi-project development
-to everyone.
+The goal of `duniverse` is to help you setup a workspace in which you only rely on dune to build
+your entire project, including its transitive dependencies.
 
-To operate, the duniverse tool relies on an external package
-manager. In version 1.0, the only supported package manager is
-opam.  The user only needs to have opam installed but not setup,
-since this tool takes care of initialising it.
+It should also make it easy for you to work on your dependencies, modify them, open PRs upstream
+or push your changes to forks to be able to reuse them while making sure none of this breaks your
+project or any of your other dependencies.
 
-Finally, the duniverse tool is tightly integrated with git. In
-particular, the workspaces it creates and manages will always be
-versioned controled by git. This ensures that users can access the
-history of the workspace as well as of individual projects inside it.
+`duniverse` relies on a `duniverse.sxp` file, at the root of the project to operate. It contains
+all the information needed to gather and inspect all of your transitive dependencies.
+For each package it know its name, the upstream git repository URL, the specific git reference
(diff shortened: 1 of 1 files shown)


doc/spec.md
```diff
@@ -1,222 +1,174 @@
...
+Because `duniverse` only supports packages that are built with `dune`, it will complain if a
+dependency cannot be resolved to a version that is built with `dune`. Hopefully, to work around
+that it will also look into the [`dune-universe/opam-overlays`]() opam repository for dunified
+forks of package which haven't been ported upstream yet. When a dependency from the `opam-overlays`
+repo is picked, it will consider it's git URL as the upstream one and not the actual upstream
+package repo URL.
+
+### System constraints
+
+`duniverse` requires that your system has OCaml and `dune` installed, in the right versions.
```
avsm: We could also vendor dune -- since 1.7, it supports being in the same directory. github.com/avsm/platform has a bootstrap script that builds dune first and uses that to build the duniverse, implying (like go) that the only thing needed to build an executable is ocaml itself.  And if we really wanted to, we could also bootstrap OCaml of course :-)

doc/spec.md
```diff
@@ -1,222 +1,174 @@
...
+field properly set to `"git+https://github.com/andre/some-important-library#v0.8.0+dune"`.
+Because the `dune-universe` team is so committed, the PR is reviewed and merged within the hour.
 
-So for instance, all the following scenarios are equivalent:
+He runs `duniverse init` once again and gets a message telling him everything went smoothly, all the
+dependencies could be resolved and the `duniverse.sxp` file was correctly written. Because André
+doesn't blindly trust any CLI tool, he runs `duniverse install` to fetch all the dependencies and
+install them in the `duniverse/` folder, move to a fresh and empty opam switch and runs
+`dune build @default @runtest` to make sure his tool builds correctly and the test suite is properly
+executed. Because the `duniverse` team did an amzing job, everything works as intended.
```
avsm: :-)

hannesm:
I like the user stories, and would appreciate some more: i.e. is the use-case that someone intends to break the API of _whatever library_ and uses duniverse (+opam) to locally clone all (most?some?) reverse dependencies in order to fix all of them supported? I happen to run into this every now and then.

NathanReb:
@hannesm yes that's indeed one of the things we want to solve with `duniverse` and I'd like to detail how in the "Mirage" user stories but we're still figuring it out atm.

I'll get back to that part of the spec soon but I need to move forward on the vendoring story first.

hannesm:
@NathanReb i was as well curious what it means to have a non-opam dependency... and why/how this is supported

NathanReb:
Hmm yeah this wasn't very clear indeed. What I meant here by a "non-opam dependency" is a source repository that build with dune, isn't necessarily packaged to oapm and that you wish to vendor.

This would be supported through a `duniverse install` command or by editing the `duniverse.sxp` file by hand. This needs to be further explained and specified here.

As to the why, I'd say to include a custom fork of an existing package for instance. One that you do not necessarily wish to publish to opam separately but want to reuse accross different projects.

avsm:
Merging this; will refine @hannesm comments about user stories in next iteration

PR #35, opened 2019-06-05
Remove GitHub variant from dev_repo

NathanReb:
Depends on #33, please do not merge before!

As promised I removed the github variant as we were not taking any advantage from it. I believe the code could probably be simplified further as I don't think the `"dev-repo without a host"` error case is necessary but let's keep things simple here. This is just another step towards the duniverse content representation!

Opening as draft until #33 is merged, I'll assign reviewers once it's done!


NathanReb:
Should be ready now!

NathanReb:
Gentle ping @Julow! Is there anything blocking?

NathanReb:
Uh good catch, thanks! Fixed it!

PR #38, opened 2019-06-11
Import stdune

NathanReb:
The need for a stdlib extension was growing so I imported `stdune` from `dune 1.10.0` into `duniverse` so we can gently start to migrate.

I'd love to do a proper PR and get rid of the extra dependencies we need to drop before merging into dune but there are more pressing matters an I think it makes sense to do that incrementally. At least new code can start using it and that's already a win!

Let me know what you think and let's try to get this merged quickly!

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,3 +1,3 @@
-(lang dune 1.6)
+(lang dune 1.8)
 (using fmt 1.1)
 (name duniverse)
--- a/lib/dune
+++ b/lib/dune
@@ -1,6 +1,6 @@
 (library
  (name duniverse_lib)
- (libraries bos ezjsonm fmt logs opam-core opam-file-format rresult sexplib
-   uri)
+ (libraries stdune bos ezjsonm fmt logs opam-core opam-file-format rresult
+   sexplib uri)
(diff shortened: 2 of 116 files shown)


NathanReb:
oh snap, ocamlformat doesn't like future_syntax :/

NathanReb:
Thank god it can be disabled in `stdune` with a:
```
disable=true
```
`stdune/.ocamlformat`! Thanks @Julow for the tip!

NathanReb:
Is there anything blocking here or anything I can do to help with the review?

NathanReb:
We do need a standard library extension at this point. If we are going to merge into `dune`, which I believe we are, I think we might as well use `stdune`, even if we don't update until the merge, it will still make it easier than migrating from another stdlib to `stdune` and it fits the purpose here.

If we're not entirely sure `duniverse` will be merge into dune, then now might the right time to figure it out because it changes our constraint quite a lot.

avsm:
I think this is a good time to try out stdune in this prototype tool.  We're pretty sure we'll merge the workflow into dune when it's right.

PR #39, opened 2019-06-14
Improve duniverse representation

NathanReb:
Depends on #38 so creating as draft.

This PR aims at using a better internal representation of the `duniverse` so that the "hard" part is building and there's no need for complex logic when manipulating it afterwards.

In other word once `duniverse init` has written the `dune-get`, rest is easy.

It's a pretty big PR compared to what we had so far so I'll try to summarize the changes here a bit. This is a refactoring and shouldn't result in a change of behaviour of the tool (with one exception that I'll bring up later).

The duniverse is now split in two parts: the configuration and the dependencies.

The configuration is just the set of parameters we use to resolve the dependencies during the `init` command and that we might eventually reuse elsewhere, especially if we need to run opam operations.

The dependencies are as the name suggest the set of dependencies for the root packages of your project. It is now split in two complementary parts: a `duniverse` which contains everything we can clone and build from sources in the `duniverse/` folder and an `opamverse` which contains all the packages we have to install through opam.
We didn't lose any information compared to what we use to have since the new representation contains the list of provided opam packages for each source repository in the `duniverse`.

We lost the log that was warning users that conflicting tag/branch were found for a given repo and telling them which one we were picking but it should be fairly easy to bring back if we really want it, which I assume we kind of do but wanted your opinion.

There's been a bunch of small refactors on the way, a lot of new tests, a bit of documentation and so on. It's just a first step but I think overall it's a huge improvement and should make it much easier to work on new features and improve the tool from there.

### DIFF
--- a/lib/duniverse.ml
+++ b/lib/duniverse.ml
@@ -1,13 +1,81 @@
+open Stdune
 open Sexplib.Conv
+
+module Element = struct
+  type t =
+    | Opam of { name : string; version : string option [@default None] [@sexp_drop_default] }
+    | Repo of
+        { dir : string;
+          upstream : string;
+          ref : string [@default "master"] [@sexp_drop_default]
+        }
+  [@@deriving sexp]
+
+  let equal t t' =
+    match (t, t') with
+    | Opam { name; version }, Opam { name = name'; version = version' } ->
+        String.equal name name' && Option.equal String.equal version version'
+    | Repo { dir; upstream; ref }, Repo { dir = dir'; upstream = upstream'; ref = ref' } ->
+        String.equal dir dir' && String.equal upstream upstream' && String.equal ref ref'
+    | (Opam _ | Repo _), _ -> false
+
+  let pp fmt t =
+    let pp_version fmt = function
+      | None -> Format.fprintf fmt "None"
+      | Some v -> Format.fprintf fmt "Some %S" v
+    in
+    match t with
+    | Opam { name; version } ->
+        Format.fprintf fmt "@[<hov 2>{ name = %S;@ version = %a }@]" name pp_version version
+    | Repo { dir; upstream; ref } ->
+        Format.fprintf fmt "@[<hov 2>{ dir = %S;@ upstream = %S;@ ref = %S }@]" dir upstream ref
+
+  let from_opam_entry ~get_default_branch entry =
+    let open Types.Opam in
+    let open Rresult.R.Infix in
+    match entry with
+    | { dev_repo = `Virtual; _ } | { dev_repo = `Error _; _ } -> Ok None
+    | { is_dune = false; package = { name; version }; _ } -> Ok (Some (Opam { name; version }))
+    | { is_dune = true; dev_repo = `Git upstream; tag = Some ref; package = { name; _ } } ->
+        Ok (Some (Repo { dir = name; upstream; ref }))
+    | { is_dune = true; dev_repo = `Git upstream; tag = None; package = { name; _ } } ->
+        get_default_branch upstream >>= fun ref -> Ok (Some (Repo { dir = name; upstream; ref }))
+
+  let dedup_upstream l =
+    let merge_repo t t' =
+      match[@warning "-4"] (t, t') with
+      | Repo { dir; ref; upstream }, Repo { dir = dir'; ref = ref'; _ } ->
+          let min_dir = match String.compare dir dir' with Lt | Eq -> dir | Gt -> dir' in
+          let max_ref =
+            match Ordering.of_int (OpamVersionCompare.compare ref ref') with
+            | Gt | Eq -> ref
+            | Lt -> ref'
+          in
+          Repo { dir = min_dir; ref = max_ref; upstream }
+      | _ -> assert false
+    in
+    let update map upstream t =
+      String.Map.update map upstream ~f:(function
+        | None -> Some t
+        | Some t' -> Some (merge_repo t t') )
+    in
+    let go (opams, upstream_repo_map) t =
+      match t with
+      | Opam _ -> (t :: opams, upstream_repo_map)
+      | Repo { upstream; _ } -> (opams, update upstream_repo_map upstream t)
+    in
+    let opams, upstream_repo_map = List.fold_left l ~init:([], String.Map.empty) ~f:go in
+    let repos = String.Map.values upstream_repo_map in
+    List.rev_append opams repos
+end
 
 type t = {
   root_packages : Types.Opam.package list;
   excludes : Types.Opam.package list;
   pins : Types.Opam.pin list;
-  packages : Types.Opam.entry list;
-  repos : Types.Dune.repo list;
   remotes : string list; [@default []]
-  branch : string [@default "master"]
+  branch : string; [@default "master"]
+  content : Element.t list
 }
 [@@deriving sexp]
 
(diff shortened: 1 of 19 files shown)


NathanReb:
Note that once #38 is merged the diff will be much smaller.

lib/types.ml
```diff
@@ -78,22 +78,3 @@ module Opam = struct
 
   let sort_uniq l = List.sort_uniq (fun a b -> String.compare a.name b.name) l
 end
```
Julow: Wouldn't it make sense to move part of the `Opam` module to `Duniverse` and remove this file?

What is in `Types`:
- `Types.Opam.entry`, `package` and `pin` are used in `Duniverse` in type declarations and be defined there.
- `Types.Opam.t` is not used anywhere and should be removed.
- You added a second module named `Opam` and that may be confusing
- `Types.Opam,Remote` is only used to represent an argument of `Exec.init_opam_and_remotes` and should be defined as a tuple there.
- `load` and `save` functions are not used anymore.
- `sort_uniq` function does not really make sense and should be moved to where it is used.

lib/duniverse.mli
```diff
@@ -1,13 +1,87 @@
...
+    (** Type of dependencies to clone in the duniverse *)
+    type t = { dir : string; upstream : string; ref : string; provided_packages : Opam.t list }
+
+    val equal : t -> t -> bool
+
+    (**/**)
+
+    (* Exposed for test purposes only *)
+
+    val raw_pp : t Fmt.t
```
Julow: `raw_pp` because it formats using ocaml syntax? Writing printers by hand seems cumbersome. Using `[@@deriving sexp]` is much easier and avoid adding bugs.

lib/duniverse.ml
```diff
@@ -1,15 +1,148 @@
...
+      t.opamverse (list Source.raw_pp) t.duniverse
+
+  let from_opam_entries ~get_default_branch entries =
+    let open Result.O in
+    let results = List.map ~f:(One.from_opam_entry ~get_default_branch) entries in
+    Result.List.all results >>= fun dep_options ->
+    let deps = List.filter_opt dep_options in
+    let opamverse, source_deps = One.partition_list deps in
+    let duniverse = Source.aggregate_packages source_deps in
+    Ok { opamverse; duniverse }
```
Julow: What about moving this 3 last lines to a `from_ones` (or `from_classified`) and nesting `One.partition_list` and `Source.aggregate_packages` ?

To split better the classify logic and the pure construction of the structure. Also, these functions do not really fit in their modules.

lib/duniverse.ml
```diff
@@ -1,15 +1,148 @@
...
+    List.equal Opam.equal t.opamverse t'.opamverse
+    && List.equal Source.equal t.duniverse t'.duniverse
+
+  let raw_pp fmt t =
+    let open Pp_combinators.Ocaml in
+    Format.fprintf fmt "@[<hov 2>{ opamverse = %a;@ duniverse = %a}@]" (list Opam.raw_pp)
+      t.opamverse (list Source.raw_pp) t.duniverse
+
+  let from_opam_entries ~get_default_branch entries =
+    let open Result.O in
```
Julow: This is usually opened globally. (When we were using `Rresult`)

lib/duniverse.ml
```diff
@@ -1,15 +1,148 @@
...
+      let update map ({ Package.upstream; _ } as package) =
+        String.Map.update map upstream ~f:(function
+          | None -> Some (from_package package)
+          | Some t -> Some (aggregate t package) )
+      in
+      let aggregated_map = List.fold_left ~init:String.Map.empty ~f:update l in
+      String.Map.values aggregated_map
+  end
+
+  module One = struct
```
Julow: Why `One` ? What about a more explicit name like `Classify` ?

lib/duniverse.mli
avsm: trying to avoid using ppx here I think -- we can't when its merged into dune

lib/pp_combinators.mli
```diff
@@ -0,0 +1,16 @@
+module Ocaml : sig
+  (** Combinators to pretty print OCaml values *)
```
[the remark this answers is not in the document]
NathanReb: Yep that's what I meant, I'll try to clarify that!

lib/pp_combinators.ml
```diff
@@ -0,0 +1,15 @@
...
+    | None -> Format.fprintf fmt "None"
+    | Some a when brackets -> Format.fprintf fmt "Some (%a)" pp_a a
+    | Some a -> Format.fprintf fmt "Some %a" pp_a a
+
+  let list pp_a fmt = function
+    | [] -> Format.fprintf fmt "[]"
+    | [ a ] -> Format.fprintf fmt "[%a]" pp_a a
+    | l ->
+        let pp_sep fmt () = Format.fprintf fmt ";@ " in
+        Format.fprintf fmt "@[<hov 2>[@ %a]@]" (Format.pp_print_list ~pp_sep pp_a) l
```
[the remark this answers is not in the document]
NathanReb: I think my reasoning was to allow it to break after the `[` but it doesn;t work so well, I'll try go make it better.

lib/types.ml
NathanReb: As I mentionned in the PR description, this is just a step in a refactoring, let's try to keep things small here.

Besides that, I think some of those types are used elsewhere and I haven't spent the time to think of a better representation for those so I'd rather do that step by step if you don't mind.

lib/duniverse.mli
NathanReb: Yep we don't want to use PPX-es anymore ahead of a dune merge.

lib/duniverse.ml
NathanReb: I'm not a fond of global opens which I often find confusing and I'm trying to get rid of those. I agree since we only use the result monad this seems harmless now but it might not always be the case and I'm not sure it's worth making an exception for this one.

lib/duniverse.ml
NathanReb: Yeah `One` is a really bad name, I'll think of a better one!

lib/duniverse.ml
NathanReb: Hmm yeah I think I see what you mean, I'll give it a go!

lib/types.ml
NathanReb: To clarify I do believe it would make sense to refine those and move them elsewhere indeed but probably as part of later refactoring!

NathanReb:
@Julow should be good now!

lib/types.ml
Julow: I understand. I still think this is in the scope of this PR as you started removing `Types.Dune` and making some functions unused and you change the way the other are used. Also `Types.Opam` is used in your new representation.
Can you at least remove the things you are making unused by this PR ? (`Types.Opam.t`, `load`, `save`)

NathanReb: Hmm yeah right, I'll remove the unused stuff, my bad!

lib/duniverse.ml
Julow: Works for me.

lib/duniverse.ml
Julow: `Classified` is great.

NathanReb:
Great! Let's merge!

PR #41, opened 2019-07-01
 Mark duniverse folder content as vendored for dune

NathanReb:
This is uses the new `vendored_dirs` stanza from dune to mark the directories in the duniverse as containing vendored code and thus prevent unwanted installable to be built and warnings to get in the way of building your own targets.

This is a draft as the corresponding stanza hasn't been merged in `dune` yet, see https://github.com/ocaml/dune/pull/2318.

It also depends on #40.


avsm:
LGTM, should we also check that the dune-project is also at least `(dune lang 1.11)`?  I had to manually update it when running duniverse on ocamlformat, for example.

cli/pull.ml
```diff
@@ -1,4 +1,35 @@
...
+open Duniverse_lib
+
+let mark_duniverse_content_as_vendored ~duniverse_dir =
+  let open Result.O in
+  let dune_file = Fpath.(duniverse_dir / "dune") in
+  let content = "(vendored_dirs *)" in
+  Logs.debug (fun l -> l "Writing %s to %a" content Styled_pp.path dune_file);
+  Bos.OS.File.write dune_file content >>= fun () ->
+  Logs.debug (fun l -> l "Successfully wrote %a" Styled_pp.path dune_file);
+  Ok ()
```
avsm: Perhaps just add this stanza rather than erase an existing `dune` file in duniverse/ ?  There might be other custom stanzas in there.

NathanReb: Do you have an example of what one would want in a `duniverse/dune`? Might be worth doing it through `duniverse` directly, eventually via CLI fags.

Since `duniverse/` is generated I think it's okay to overwrite whatever is in there, at least in a first iteration and especially since blindly appending could lead to weird stuff as well.

We could also add a comment saying this file is generated and discourage editing it.

I think while it's delusional to think we'll get it right in this PR straight away and that people won't need to edit the content of the `duniverse/`, it definitely is what we should aim for!

NathanReb:
Yeah I agree the language version thing is annoying, I'll add that with a little prompt (and a `-y` flag)  as I'm not to keen on editing user files without warning.

cli/pull.ml
avsm: I mainly just have local hacks in there at the moment (e.g.  a rule to copy executables into an output/ directory).  Let's go with your suggestion of just adding a comment to mark it as autogenerated... I totally agree with your sentiment that duniverse should "just work" :-)

avsm:
Sounds good!

avsm:
Merging now that https://github.com/ocaml/dune/pull/2318 is in dune master!

NathanReb:
Oops, I need to remove a commit where I pinned the dune dependency, I'll do that on master!

PR #43, opened 2019-07-08
Unpin dune

NathanReb:
We should probably wait until 1.11 is released to merge that but it should be any day now!

### DIFF
--- a/duniverse.opam
+++ b/duniverse.opam
@@ -23,7 +23,7 @@ depends: [
   "alcotest" {with-test}
 ]
 pin-depends: [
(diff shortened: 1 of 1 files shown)


duniverse.opam
```diff
@@ -22,9 +22,6 @@ depends: [
   "opam-core"
   "alcotest" {with-test}
 ]
-pin-depends: [
-  ["dune.dev" "git+https://github.com/NathanReb/dune.git#vendor-stanza"]
```
avsm: Could you just point this to "git+https://github.com/dune/dune.git for now? That'll keep CI working until the release

PR #45, opened 2019-07-10
opam file generation

avsm:


NathanReb:
Looks good to me!

I'll format it and merge, thanks!

NathanReb:
Hmm the issue comes from the language version. Any particular reason why you updated it to 1.11?

NathanReb:
I need to make the ocamlformat CI build more robust!

NathanReb:
Updated the ocamlformat docker images which fixed the CI. Merging, thanks!

NathanReb:
I pushed a few commit to have the CI check that `dune-project` and `duniverse.opam` are in sync.

Here's an example of how it looks when they're not:
Screenshot from 2019-07-10 15-09-07

If that suits you I think we can merge!

PR #49, opened 2019-07-11
Resolve git refs to commit hashes

NathanReb:
Depends on #48

This adds a step in `duniverse init` that resolves git refs to actual commit hashes before writing the duniverse.

This is still WIP as it requires a couple follow up steps:
1. [x] Actually checkout to the said commit hash in `duniverse pull`
2. [x] Add a `duniverse update` command to "refresh" commit hashes


NathanReb:
Also I'm not super convinced by the "polymorphic Deps.t" thing I did but that was the easiest way to do the resolving once source repos have been aggregated and thus to avoid running `git ls-remote` more than we actually need.

cli/pull.ml
```diff
@@ -1,3 +1,4 @@
+module Buenzli_fmt = Fmt
```
emillon: Why is this rename necessary? If it is, how about something more neutral like `Fmt_lib`?

lib/duniverse.ml
```diff
@@ -19,34 +23,37 @@ module Deps = struct
...
-      ref : string; [@default "master"] [@sexp_drop_default]
+      ref : 'ref;
       provided_packages : Opam.t list [@default []] [@sexp_drop_default]
     }
     [@@deriving sexp]
 
-    let equal t t' =
-      String.equal t.dir t'.dir && String.equal t.upstream t'.upstream && String.equal t.ref t'.ref
+    let equal equal_ref t t' =
+      String.equal t.dir t'.dir
```
emillon: I suggest using pattern matching here to make sure that it breaks when more fields are added.

lib/duniverse.ml
```diff
@@ -133,6 +146,11 @@ module Deps = struct
     classify ~get_default_branch entries >>= fun classified -> Ok (from_classified classified)
 
   let count { opamverse; duniverse } = List.length opamverse + List.length duniverse
+
+  let resolve ~resolve_ref t =
+    let open Result.O in
+    Result.List.map ~f:(Source.resolve ~resolve_ref) t.duniverse >>= fun duniverse ->
+    Ok { t with duniverse }
```
emillon: you can use map instead of bind here :)

lib/exec.ml
```diff
@@ -114,6 +117,16 @@ let git_add_all_and_commit ~repo ~message () =
...
 let git_merge ?(args = Cmd.empty) ~from ~repo () = run_git ~repo Cmd.(v "merge" %% args % from)
 
+let git_resolve ~remote ~ref =
+  run_and_log_l Cmd.(v "git" % "ls-remote" % remote % ref) >>= function
+  | [ line ] -> Git.parse_ls_remote_line line >>= fun (commit, _) -> Ok { Git.Ref.t = ref; commit }
+  | [] -> Error (`Msg (Format.asprintf "No %a ref for %s" Git.Ref.pp ref remote))
+  | _ ->
+      Error
+        (`Msg
+          (Format.asprintf "Ref %a is pointing to more than one commit for %s" Git.Ref.pp ref
```
emillon: You can use `Rresult.R.error_msgf` instead of `Error (Msg (asprintf ...))`

cli/pull.ml
NathanReb: Something more neutral would work indeed. The rename is need because `Stdune` as an `Fmt` module which is basically `Fmt` shrinked to the bare minimum.

lib/duniverse.ml
NathanReb: Yes good idea!

lib/exec.ml
NathanReb: True, I always forget about that one!

lib/duniverse.ml
NathanReb: It is in fact intentional!

I find it more annoying and less diff friendly to have to update the infix operator when moving those lines around or further binding the result.

I agree in this case since there's only this one it would make sense but I prefer to adopt a consistent style and only use `Result.bind`.

NathanReb:
@avsm do you want to take a look or should I go ahead and merge?

avsm:
lgtm!

PR #50, opened 2019-07-16
Rely less on dev-repo field (e.g. for opam pins)

TheLortex:
In an attempt to solve https://github.com/avsm/duniverse/issues/37
Basically it tries to parse url.src as a dev-repo field before having to fallback on the actual dev-repo field.

### DIFF
--- a/test/test_opam_cmd.ml
+++ b/test/test_opam_cmd.ml
@@ -48,10 +48,10 @@ let test_tag_from_archive =
   ]
 
 let test_classify_package =
-  let make_test ~name ~package ?(dev_repo = "dummy-dev-repo") ?archive ?(pins = []) ~expected () =
+  let make_test ~name ~package ?(dev_repo = "dummy-dev-repo") ?archive ~expected () =
     let test_name = Printf.sprintf "classify_package: %s" name in
     let test_fun () =
-      let actual = Duniverse_lib.Opam_cmd.classify_package ~package ~dev_repo ~archive ~pins () in
+      let actual = Duniverse_lib.Opam_cmd.classify_package ~package ~dev_repo ~archive () in
       Alcotest.(check (pair Testable.opam_repo (option string))) test_name expected actual
     in
     (test_name, `Quick, test_fun)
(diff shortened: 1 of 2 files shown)


avsm:
lgtm, it's nice to just keep the url.src as the one field that changes.

NathanReb:
I haven't looked at the changes yet but I'm curious to know what your motivation is for this?

It seems a bit weird to me to rely on `url.src` rather than on `dev-repo` considering that `duniverse` relies exclusively on git to pull the sources of your dependencies. `url.src` doesn't have to be a version controlled url and can point to anything, even when `dev-repo` is a github URL.

Don't get me wrong, I'm happy to merge this if it solves some issues but I think I just need a bit more context to better understand what changes exactly are required. E.g. what was the `duniverse` error you got that lead to this PR?

avsm:
The problem I see is that dev-repo points at the original one, and then fails to check out the right branch (which comes from url.src).  It seems easier to just depend on url.src if at all possible, as that's the definitive source of the archive.

TheLortex:
For example in an opam file that has a `pin-depends` field, the `url.src` might refer to a fork, and thus `dev-repo` won't be in sync.

NathanReb:
Hmm ok I see. I'm not sure I understand how `pin-depends` has any influence on `url.src` but I get the general idea.

It might be worth refactoring this bit of the code to clarify when we rely on `url.src` and when we rely `dev-repo` rather than just falling back to one when we can't parse the other but we can do that later. I'll take a deeper look, thanks for the clarification!

test/test_opam_cmd.ml
```diff
@@ -90,7 +90,16 @@ let test_classify_package =
...
+      ~dev_repo:"git+https://host.com/some-repo.git" ~archive:"git+https://host.com/some-fork.git#dev"
+      ~expected:(`Git "https://host.com/some-fork.git", Some "dev")
+      ();
+    make_test ~name:"fallback to dev_repo" ~package:{ name = "x"; version = None }
+      ~dev_repo:"git+https://host.com/some-repo.git" ~archive:"https://github.com/user/repo/releases/download/v1.2.3/archive.tbz"
+      ~expected:(`Git "https://host.com/some-repo.git", Some "v1.2.3")
       ()
   ]
 
+
```
NathanReb: A newline slipped in!

TheLortex: oop

PR #52, opened 2019-07-19
Fix dune-project according to latest opam file changes

NathanReb:
The opam file was updated instead of the `dune-project`!


avsm:
hah! and i was the one to suggest the comment on the opam file as well :-)

PR #53, opened 2019-07-19
Fix git ref resolution with packed-refs

NathanReb:
This fixes an issue with a log about the registered commit not being the head commit for the ref anymore. It used to appear when the repo had packed-refs.

The bug is fixed and that log is now info level since there are valid use case for sticking to an older commit.

### DIFF
--- a/lib/exec.ml
+++ b/lib/exec.ml
@@ -118,12 +118,11 @@ let git_add_all_and_commit ~repo ~message () =
 let git_merge ?(args = Cmd.empty) ~from ~repo () = run_git ~repo Cmd.(v "merge" %% args % from)
 
 let git_resolve ~remote ~ref =
-  run_and_log_l Cmd.(v "git" % "ls-remote" % remote % ref) >>= function
-  | [ line ] -> Git.parse_ls_remote_line line >>= fun (commit, _) -> Ok { Git.Ref.t = ref; commit }
-  | [] -> Rresult.R.error_msgf "No %a ref for %s" Git.Ref.pp ref remote
-  | _ ->
-      Rresult.R.error_msgf "Ref %a is pointing to more than one commit for %s" Git.Ref.pp ref
-        remote
+  run_and_log_l Cmd.(v "git" % "ls-remote" % remote %% Git.Ls_remote.ref_arg ref) >>= fun output ->
+  match Git.Ls_remote.commit_pointed_by ~ref output with
+  | Ok commit -> Ok { Git.Ref.t = ref; commit }
+  | Error `No_such_ref -> Rresult.R.error_msgf "No %a ref for %s" Git.Ref.pp ref remote
+  | Error (`Msg _) as err -> err
 
 let opam_cmd ~root sub_cmd =
   let open Cmd in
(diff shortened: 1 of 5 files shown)


NathanReb:
Maybe @TheLortex you care to take a look while @avsm is away so we can merge that?

lib/exec.ml
```diff
@@ -118,12 +118,11 @@ let git_add_all_and_commit ~repo ~message () =
 let git_merge ?(args = Cmd.empty) ~from ~repo () = run_git ~repo Cmd.(v "merge" %% args % from)
 
 let git_resolve ~remote ~ref =
-  run_and_log_l Cmd.(v "git" % "ls-remote" % remote % ref) >>= function
-  | [ line ] -> Git.parse_ls_remote_line line >>= fun (commit, _) -> Ok { Git.Ref.t = ref; commit }
-  | [] -> Rresult.R.error_msgf "No %a ref for %s" Git.Ref.pp ref remote
-  | _ ->
-      Rresult.R.error_msgf "Ref %a is pointing to more than one commit for %s" Git.Ref.pp ref
```
TheLortex: This case is now ignored: this might be a problem, right ?

NathanReb: I doubt it can but I can bring it back to life if needed. Do you think this can actually happen?

TheLortex: It might happen if a branch and a tag have the same name

NathanReb: Good point, I'll fix that!

NathanReb: I came with a quick fix. It's not perfect but it will do for now.

I think in the future we need to make it explicit whether we track a tag or a branch and eventually prompt the user when both are available and we have to choose.

NathanReb:
Does that look good to you @TheLortex ? May I go ahead and merge?

lib/git.ml
```diff
@@ -1,10 +1,45 @@
...
+    let is_ref full_ref = String.is_suffix ~suffix:ref full_ref in
+    let is_non_packed_ref full_ref = String.is_suffix ~suffix:non_packed full_ref in
+    let rec go acc lines =
+      match (acc, lines) with
+      | _, [] -> Ok acc
+      | { not_packed = None; _ }, (commit, full_ref) :: tl when is_non_packed_ref full_ref ->
+          go { acc with not_packed = Some commit } tl
+      | { maybe_packed = None; _ }, (commit, full_ref) :: tl when is_ref full_ref ->
+          go { acc with maybe_packed = Some commit } tl
+      | _, (_, full_ref) :: _ when is_non_packed_ref full_ref || is_ref full_ref ->
```
TheLortex: Maybe this shouldn't fail if both refs point to the same commit.

NathanReb: I think it should because if the ref is both a tag and branch, it might point to different commits in the future.

It would also need quite a bit of complex logic to deal with all possible cases with packed refs.

As I said above, I agree this isn't great but it wasn't any better before. I think the proper fix is to explicitly specify whether to track a branch or a tag, in which case this ambiguity doesn't exist anymore. I'm happy to do that in a separate PR as I believe this already fixes a bug in the current implementation and there's value in the PR as it is.

NathanReb: We could improve the error message though and change it to `"Several refs share this name"` or something similar.

TheLortex: Okay then I'm fine with that !

NathanReb: I'll change the message, I think it's clearer!

PR #54, opened 2019-07-26
Use threads to speed up init and pull

TheLortex:
`duniverse init` and `duniverse pull` spend a great amount of time waiting for the network. Using threads allows a nice amount of parallelization, thus improving overall speed.

Some benchmarks for 105 packages.

`duniverse init`: 
* 1 thread: ∞ (> 1m)
* 20 threads: 30s
* 100 threads: 28s

`duniverse pull`:
* 1 thread: ∞ (> 1m)
* 24 threads: 10s

### DIFF
--- a/cli/common.ml
+++ b/cli/common.ml
@@ -9,12 +9,23 @@ module Arg = struct
     let doc = "Do not prompt for confirmation and always assume yes" in
     Cmdliner.Arg.(value & flag & info [ "y"; "yes" ] ~doc)
 
+  let thread_safe_reporter reporter =
+    let lock = Mutex.create () in
+    let { Logs.report } = reporter in
+    let oui src level ~over k msgf =
+      Mutex.lock lock;
+      let x = report src level ~over k msgf in
+      Mutex.unlock lock;
+      x
+    in
+    Logs.{ report = oui }
+
   let setup_logs () =
     Printexc.record_backtrace true;
     let setup_log style_renderer level =
       Fmt_tty.setup_std_outputs ?style_renderer ();
       Logs.set_level level;
-      Logs.set_reporter (Logs_fmt.reporter ())
+      Logs.set_reporter (thread_safe_reporter (Logs_fmt.reporter ()))
     in
     let global_option_section = "COMMON OPTIONS" in
     let open Cmdliner.Term in
--- a/cli/pull.ml
+++ b/cli/pull.ml
@@ -70,15 +70,12 @@ let warn_about_head_commit ~ref ~commit () =
   Common.Logs.app (fun l -> l "You might want to consider running 'duniverse update'");
   ()
 
-let checkout_if_needed ~head_commit ~output_dir ~ref ~commit ~dir () =
+let checkout_if_needed ~head_commit ~output_dir ~ref ~commit () =
   let open Result.O in
   if String.equal commit head_commit then Ok ()
   else (
     warn_about_head_commit ~ref ~commit ();
-    Exec.git_unshallow ~repo:output_dir () >>= fun () ->
-    match Exec.git_checkout ~repo:output_dir commit with
-    | Ok () -> Ok ()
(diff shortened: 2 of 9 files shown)


cli/pull.ml
```diff
@@ -89,7 +89,7 @@ let pull ~duniverse_dir src_dep =
   Bos.OS.Dir.delete ~recurse:true output_dir >>= fun () ->
   Exec.git_shallow_clone ~output_dir ~remote:upstream ~ref () >>= fun () ->
   Exec.git_rev_parse ~repo:output_dir ~ref:"HEAD" () >>= fun head_commit ->
-  checkout_if_needed ~head_commit ~output_dir ~ref ~commit () >>= fun () ->
+  checkout_if_needed ~head_commit ~output_dir ~ref ~commit ~dir () >>= fun () ->
```
Julow: To avoid passing an unneeded argument, you could use ``|> R.reword_error (fun (`Msg _) -> `Commit_is_gone dir)`` here.

lib/parallel.ml
```diff
@@ -0,0 +1,31 @@
...
+        worker result
+  in
+  let threads, results =
+    List.init n_threads ~f:(fun _ ->
+        let result = ref [] in
+        (Thread.create worker result, result) )
+    |> List.split
+  in
+  List.iter ~f:Thread.join threads;
+  List.map ~f:( ! ) results |> List.flatten
```
Julow: You could do the `split`, `iter`, `map` and `flatten` in a single `fold` but it would be slightly less readable.

lib/parallel.ml
```diff
@@ -0,0 +1,31 @@
...
+  let rec worker result =
+    Mutex.lock queue_lock;
+    let hd =
+      match !queue with
+      | [] -> None
+      | hd :: tl ->
+          queue := tl;
+          Some hd
+    in
+    Mutex.unlock queue_lock;
```
Julow: The protected queue code could be moved out of the function for readability. (eg. `seq`-style `protected_queue : 'a list -> (unit -> 'a option)`)

cli/common.ml
```diff
@@ -9,12 +9,23 @@ module Arg = struct
...
+  let thread_safe_reporter reporter =
+    let lock = Mutex.create () in
+    let { Logs.report } = reporter in
+    let oui src level ~over k msgf =
+      Mutex.lock lock;
+      let x = report src level ~over k msgf in
+      Mutex.unlock lock;
+      x
+    in
+    Logs.{ report = oui }
```
Julow: We should try to upstream that.

TheLortex:
Updated according to @Julow and @NathanReb reviews !
> Also, it might be worth doing the same thing for git ref resolution, both in init and update but that can wait!
It's already done for `init` ref resolution but `update` needs a little bit more work as it uses an `incr` on a shared `ref` so I'm not sure to what extent it's thread safe.

test/test_parallel.ml
```diff
@@ -0,0 +1,22 @@
+let test_parallel =
+  let make_test ~name ~input ~f ~expected () =
+    let test_name = Printf.sprintf "parallel: %s" name in
```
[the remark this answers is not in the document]
TheLortex: Indeed

NathanReb:
Thanks! :rocket: :roller_coaster: :speedboat: :racehorse: :racing_car:

lib/parallel.ml
```diff
@@ -0,0 +1,37 @@
...
+        result := f hd :: !result;
+        worker result
+  in
+  List.init n_threads ~f:(fun _ ->
+      let result = ref [] in
+      (Thread.create worker result, result) )
+  |> List.fold_left
+       ~f:(fun acc (thread, result_ref) ->
+         Thread.join thread;
+         !result_ref @ acc )
```
Julow: `List.rev_append` should be faster and avoid shuffling the list (the order is already lost so this don't matter)

PR #55, opened 2019-07-30
CI fixes

NathanReb:


dune
```diff
@@ -8,3 +8,5 @@
  (mode promote)
  (action
   (copy doc/usage.md %{targets})))
+
+(vendored_dirs stdune)
```
emillon: I think you'll have to bump the `dune` lang & dependency in opam.

NathanReb: It was dune `1.11` already!

emillon: ah nevermind I saw an example `opam` file in the diff...

PR #59, opened 2019-08-16
self host duniverse!

avsm:


avsm:
This is now ready for merge.  One hitch is that the duniverse'd up repository doesn't build with dune-master, due to the `dune.configurator` builtin library not being found.

```
$ make
dune build 
File "duniverse/base/src/discover/dune", line 1, characters 41-58:
1 | (executables (names discover) (libraries dune.configurator)
                                             ^^^^^^^^^^^^^^^^^
Error: Library "dune.configurator" not found.
Hint: try: dune external-lib-deps --missing @@default
Makefile:6: recipe for target 'build' failed
```

I've filed https://github.com/ocaml/dune/issues/2679 upstream to figure this out, but the current tree is mergeable with dune 1.11.3

avsm:
Note:

- I've disabled the OCAMLFORMAT tests, as they break with the dune file generated the duniverse. I don't think its worth bringing ocamlformat back until the tool is more stable.
- The CI now builds with a `git clone; dune build` and then does the opam installation.

NathanReb:
Can you point to one of the failures with ocamlformat? If it's easy enough to fix I'd like to keep that check!

NathanReb:
Another remark I have is that now that we vendor dependencies it might make more sense to remove them from the opam file.

I think it'll be hard to keep valid constraints there since opam isn't part of the duniverse development workflow anymore. I agree we can't do it now as we don't have a pure duniverse way of upgrading to the latest opam version but I think we should try to integrate that ASAP!

What do you think?

PR #62, opened 2019-08-21
Improve parsing of git ls-remote

Julow:
See https://github.com/dune-universe/opam-overlays/pull/49#issuecomment-522923531

- Only search for `refs/heads/<ref>` (branches) and `refs/tags/<ref>` (tags) in the output of `git ls-remote`.

Remotes may have more refs but they should/can not be used, we only want branch/tags.
There should never be two refs with the same full name (`refs/...`).

- Fix a bug with the `is_suffix` logic

"abc" could match "refs/heads/xabc"

- ~~Remove the `Multiple_such_ref` error and choose tags first~~

~~If a tag and a branch have the same name, the tag is chosen. To avoid failing.~~

### DIFF
--- a/lib/exec.ml
+++ b/lib/exec.ml
@@ -122,6 +122,8 @@ let git_resolve ~remote ~ref =
   match Git.Ls_remote.commit_pointed_by ~ref output with
   | Ok commit -> Ok { Git.Ref.t = ref; commit }
   | Error `No_such_ref -> Rresult.R.error_msgf "No %a ref for %s" Git.Ref.pp ref remote
+  | Error `Multiple_such_refs ->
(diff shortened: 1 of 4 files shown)


NathanReb:
I haven't looked at the changes yet but we shouldn't pick tags over branches. If we want to avoid failures there we can eventually add a prompt but we should definitely offer a way for users to decide which one they want.

We could eventually pick tags over branches when we know we inferred the tag from the distribution archive for an opam package hosted on github but even I'm not sure it's worth having to convey that information around.

Julow:
To avoid ambiguity I reimplemented the `Multiple_such_refs` error. This PR still fixes the two issues.

NathanReb:
Can you please just rebase and format so we can merge?

Julow:
Done.

NathanReb:
Thank you very much!

PR #63, opened 2019-08-22
Path pin local packages when initializing duniverse

NathanReb:
A few users have been reporting issues when working with duniverse when the opam file contains uncommitted changes. This I believe it is due to the fact that opam infers the pinning method based on the content of the directory and picks `"git"`. Now we explicitly pass `path` when pinning the local packages so that `duniverse init` reflects the current state of the opam file which makes it much more convenient to work with.


NathanReb:
cc @dinosaure, this should fix your issue, if you care to try it!

NathanReb:
Thanks for the review! Let's merge!

PR #65, opened 2019-09-25
Improve use of cmdliner Args

NathanReb:
Cmdliner Args are now wrapped with a constructor to avoid confusion in the `run` function arguments.

@craigfe set that up for `mdx` and I thought it would be a good idea to do the same here!

### DIFF
--- a/cli/common.mli
+++ b/cli/common.mli
@@ -5,13 +5,18 @@ module Logs : sig
 end
 
 module Arg : sig
+  val named : ('a -> 'b) -> 'a Cmdliner.Term.t -> 'b Cmdliner.Term.t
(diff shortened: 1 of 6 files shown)


cli/common.mli
```diff
@@ -5,13 +5,18 @@ module Logs : sig
 end
 
 module Arg : sig
+  val named : ('a -> 'b) -> 'a Cmdliner.Term.t -> 'b Cmdliner.Term.t
```
Julow: Why putting this in `Arg` ? This module contains all the arguments (and not overloading `Cmdliner.Arg`) and this is not an argument itself. What do you think of moving it into `Common` ?

NathanReb: I think that's where it belongs, we use it to build arguments consistently across the commands.

PR #66, opened 2019-09-30
Re-enable formatting CI check

NathanReb:
This PR re-enables the formatting check by fixing the duniverse-generated dune files. I think it's even more important to check for formatting now since it will allow us to figure out if we produce badly formatted dune files and I think that's something we should avoid!

Eventually we'll use dune to format our generated dune files in the future but this made for a simpler fix!


NathanReb:
I have rebased on top of #67 to emphasize that the deps were not modified!

NathanReb:
Meaning we should merge #67 first!

PR #68, opened 2019-09-30
Allow to `duniverse update` specific repositories

NathanReb:
Depends on #67 and #67, I'll undraft once they're merged!

This adds arguments to the `duniverse update` command so that you can update the tracked commit for a subset of the duniverse! This can be pretty convenient!


avsm:
lgtm!

PR #69, opened 2019-10-01
Fix formatting from #29

NathanReb:


NathanReb:
This was before we re-enabled formatting checks in CI so I missed that one!

PR #71, opened 2019-10-24
Add explicit version to .ocamlformat

NathanReb:
I believe this is best practice although the error message isn't necessarily great it's still better than reformatting everything by mistake if you run the wrong version of ocamlformat.

It will also be helpful to benefit from https://github.com/ocaml-ci/ocaml-ci/pull/49

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,2 +1,3 @@
+version=0.9
 profile=conventional
 margin=100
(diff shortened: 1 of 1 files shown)


NathanReb:
We should probably enable `ocaml-ci` here @avsm !

NathanReb:
Upgrading ocamlformat result in a fairly big reformatting of the code base so I'd rather wait for a more stable release, especially given we haven't hit any major bug so far with `0.9`!

PR #72, opened 2019-12-02
Switch from an overlay remote to a fully materialised opam repository

avsm:
The old scheme was rather non-deterministic, since ocaml/opam-repository and dune-universe/opam-overlays would advance out of sync.

The new scheme defaults `duniverse` to using a combined `dune-universe/opam-repository#duniverse` which has everything in one place.  The idea is that we only push ocaml/opam-repository updates here once it passes CI, so `duniverse init` should keep working!


avsm:
This is working pretty well locally now; I have some associated fixes to ocaml-ci to make it work out of the box with this setup as well.

avsm:
Merging this to break the CI cycle - it restores dune 2 compat

PR #74, opened 2020-01-12
Pull source as git submodules by default

avsm:
`duniverse pull` will now clone the source code and add each of
the vendored packages as a git submodule. It does this by adding
the submodule commit objects after the source pulls have been done.
This is a lot faster than the default `git submodule add` which
requires setting up remote tracking branches for each of the
remotes.

This adds a `--no-submodules` option to `pull` to just do a
source code clone.

### DIFF
--- a/cli/pull.ml
+++ b/cli/pull.ml
@@ -146,29 +146,27 @@ let pull_source_dependencies ?trim_clone ~duniverse_dir ~cache src_deps =
 
 let get_cache ~no_cache = if no_cache then Ok Cloner.no_cache else Cloner.get_cache ()
 
-let run (`Yes yes) (`No_cache no_cache) (`No_submodules no_sm) (`Repo repo) () =
+let run (`Yes yes) (`No_cache no_cache) (`Repo repo) () =
   let open Result.O in
   let duniverse_file = Fpath.(repo // Config.duniverse_file) in
   Duniverse.load ~file:duniverse_file >>= function
   | { deps = { duniverse = []; _ }; _ } ->
       Common.Logs.app (fun l -> l "No dependencies to pull, there's nothing to be done here!");
       Ok ()
-  | { deps = { duniverse; _ }; _ } ->
+  | { deps = { duniverse; _ }; config } ->
+      let sm = Duniverse.Config.(config.pull_mode = Submodules) in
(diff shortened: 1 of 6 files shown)


cli/pull.ml
```diff
@@ -125,12 +158,18 @@ let run (`Yes yes) (`No_cache no_cache) (`Repo repo) () =
...
+      pull_source_dependencies ~trim_clone:(not no_sm) ~duniverse_dir ~cache duniverse >>= fun () ->
+      if no_sm then Ok () else set_git_submodules ~repo ~duniverse_dir duniverse
 
 let no_cache =
   let doc = "Run without using the duniverse global cache" in
   Common.Arg.named (fun x -> `No_cache x) Cmdliner.Arg.(value & flag & info ~doc [ "no-cache" ])
 
+let no_submodules =
+  let doc = "Run without adding the source as submodules" in
+  Common.Arg.named (fun x -> `No_submodules x) Cmdliner.Arg.(value & flag & info ~doc [ "no-submodules" ])
```
Julow: Cmdliner already adds a `--no-` for flags, this adds `--no-submodules` and `--no-no-submodules`.

avsm:
It can't be on `init`, since the submodule generation requires pulling the source trees. Git submodules work by generating a special git link object based on the commit ref of the checked out submodule, and so they can't work without a checkout of the individual modules.

I'm thinking about the common case use here -- most beta testers of duniverse have been fairly clear that they don't want to check in the source code directly, so git submodules are the next best thing.  Your point about messing with the git history is important: however, this command can be reverted by simply `git reset --hard` as it only stages the files. 

There is also an important missing piece of this PR: to delete unused gitmodules by comparing the present modules to the requested ones. This implies duniverse initially will not support projects with other external submodules, but this seems an acceptable limitation for a first release.

Julow:
I meant that the option is passed to `init` to be stored in the `dune-get` file, so the user doesn't have to pass this option on every `pull`. Why is that not possible ?

avsm:
That makes sense.  I'm happy with submodules being the default recorded value on `init`, since that's the workflow we want to promote.

cli/pull.ml
avsm: flag removed

avsm:
Pushed the change so that the pull_mode is stored in dune-get. Thanks for the suggestion @Julow!

avsm:
@samoht this is ready to try out if you want

PR #76, opened 2020-02-12
Allow pulling a subset of dependencies

NathanReb:
This allows running `duniverse pull repo1 repo2 ...` the same way we do for `duniverse update`. It can be quite convenient, in particular in RWO where I tend to update only mdx quite often. It's especially annoying there where pulling other deps requires me to run `dune upgrade --root duniverse/<jbuilder_repo>`.

In the long run we probably want to make that dune upgrade part of the pulling process but in the meantime this is enough.


NathanReb:
That's a good point. We have the same issue in `duniverse update` actually. I'll fix that!

For the version thing, my guess is that the versions in the duniverse dirs are there because at some point we hope to be able to have several of them cohabitating here, in which case we do want the version number. If I'm wrong then I'm happy to make it match on the provided packages instead!

NathanReb:
It's worth noting that this can be done in a subsequent PR as having this around for RWO is quite convenient.

NathanReb:
I pushed some changes:
- Delete existing dirs before pulling, I ran into this and it was causing the whole thing to fail with a quite unhelpful error message!
- Check that all CLI provided repos match repos in the duniverse (for `pull` and `update`)

Let me know what you think!

That `diff` function could probably use some test but I was unsure where to extract it for now and it's easier to implement a version specific to this use case.

cli/common.ml
```diff
@@ -43,3 +60,37 @@ module Logs = struct
...
+  List.mem ~set:to_consider src_dep.dir
+
+(** Filters the duniverse according to the CLI provided list of repos *)
+let filter_duniverse ~to_consider src_deps =
+  let open Rresult in
+  match to_consider with
+  | None -> Ok src_deps
+  | Some to_consider ->
+      let filtered = List.filter ~f:(should_consider ~to_consider) src_deps in
+      check_duniverse_repos to_consider filtered >>= fun () -> Ok filtered
```
Julow: This is a lot of code ! I think all this can be simplified with an assumptions: the result correspond to `to_consider` (it's not filtered). The only needed logic is to lookup in `src_deps` (as a map):
```ocaml

(** Filters the duniverse according to the CLI provided list of repos *)
let filter_duniverse ~to_consider src_deps =
  let open Rresult in
  match to_consider with
  | None -> Ok src_deps
  | Some to_consider -> (
      let found =
        let f (src : _ Duniverse.Deps.Source.t) = (src.dir, src) in
        String.Map.of_list_map_exn src_deps ~f
      in
      let unmatched, to_consider =
        List.partition_map to_consider ~f:(fun asked ->
            match String.Map.find found asked with None -> Left asked | Some src -> Right src)
      in
      match (unmatched, to_consider) with
      | [], to_consider -> Ok to_consider
      | unmatched, _ ->
          let sep fmt () = Ffmt.pf fmt " " in
          Rresult.R.error_msgf "The following repos are not in your duniverse: %a"
            Ffmt.(list ~sep string)
            unmatched )
```

This example uses functions from stdune. (which is in scope because of dune-configurator and used in every modules)

NathanReb: You're right this looks much better! I'm not sure what I had in mind when I wrote that :')

avsm:
Looks good to me. I'll merge it after #79 has gone in, as I'd like to get that working in opam-ci as a high priority.

NathanReb:
@avsm please let me implement @Julow 's suggestion first!

avsm:
Sounds good.

NathanReb:
Done!

avsm:
Resolved conflicts.

NathanReb:
@avsm is this good to merge?

avsm:
Thanks!

PR #87, opened 2020-06-05
Add pin management to duniverse

rizo:
This PR adds support for pinned packages to duniverse.

Pinned packages can be added to the `dune-get` file with `duniverse pin PACKAGE REPO`. When a pin is found in `dune-get` during `duniverse init` it will be pulled and included in the dependency resolution. An opam entry with a fixed ref is attributed to each pin and saved in `dune-get` to achieve reproducible pulls and builds.

In addition to including the pins in the dependency resolution, the opam files of the pinned packages are stored in `Config.pins_dir`. These files can be optionally changed to override build options or dependencies, for example.

## Duniverse file changes

The pinned packages are included in the `config` section of the duniverse. Here's an example:

```lisp
((config
  ((version 1) (root_packages (((name my-app))))
   (pins
    (((pin coap-server-lwt)
      (url (https://github.com/hypercollective/ocaml-coap.git)))
     ((pin coap-core)
      (url (https://github.com/hypercollective/ocaml-coap.git)))))
...
```

## duniverse pin/unpin

Two convenience commands are added to help updating the `dune-get` file:

```
$ duniverse pin coap-core git+https://github.com/hypercollective/ocaml-coap.git
==> Added pin coap-core to dune-get. You can now run duniverse init to update the dependencies.

$ duniverse unpin coap-core
==> Removed pin coap-core from dune-get. You can now run duniverse init to update the dependencies.
```

Currently these commands **do not** automatically update the dependencies. The initial motivation is that users might want to add/remove multiple pins, but I'm actually thinking about changing this to force (re)`init` after each pin/unpin.

## Notes
1. Running `duniverse init` when pins are present in the duniverse file, will pull the sources into the duniverse folder. This is needed because we need to get the opam files for the pinned packages (so that dependencies can be correctly resolved).
    - I have thought about pulling pins into a temporary folder, but honestly, I'm not sure if it matters that much.
2. The obtained opam files, after they're used for dependency resolution, are also put in `./duniverse/.pins`.
    - We want to allow users to change those files and re-run `duniverse init` to update any `depends` changes.
    - When running `duniverse init`, existing opam files in `./duniverse/.pins` will not be overwritten.
    - Manually copying opam files into `./duniverse/.pins` is **not** sufficient for `init` to work properly, they must also be present in `dune-get`.
    - If a pin is removed from `dune-get`, running `duniverse init` will remove the associated opam file.
    - Currently the pins directory is `./duniverse/.pins`. Using a hidden path is important because this way dune will not attempt to analyse the opam files and complain about duplicate packages. We should probably rename this to `./duniverse/_pins` as using hidden paths might be confusing (see https://dune.readthedocs.io/en/stable/dune-files.html#dirs-since-1-6).
3. There might be some problems if a pinned opam file has a `url.src` or `dev-repo` fields. I'll submit a commit to address this soon.
4. Even though a new field was added to `config` in `dune-get`, the change is backwards compatible. The `pins` field is optional.
5. It is assumed that pinned packages are dune-compatible. I plan on improving error reporting, in case they're not.

## Final thoughts

I recommend reading individual commits in the chronological order to understand this PR.
One particular implementation detail is the fact that pins are fetched during init as a "mini-duniverse". This is very convenient because we can benefit from existing logic to ensure fixed refs, and related pins are automatically grouped.

Also many thanks to @avsm for suggestions about how to approach this feature! 🙌  This is probably still incomplete, but I'm already using this in a private project and am planning on improving things as I run into issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,11 +52,16 @@ COMMANDS
        opam-install
            install packages that are not duniverse-compatible via opam
 
+       pin Add a pinned package dependency to dune-get
+
        print-ocaml-compilers
            print OCaml compilers that are supported for this duniverse
 
        pull
            fetch the latest archives of the vendored libraries
+
+       unpin
+           Remove a pinned package dependency from dune-get
 
        update
            update the commit hash corresponding to the tracked branch/tag for
--- a/bin/duniverse.ml
+++ b/bin/duniverse.ml
@@ -17,6 +17,7 @@
 open Duniverse_cli
 
 let cmds =
-  [ Init.cmd; Update.cmd; Pull.cmd; Opam_install.cmd; Depext.cmd; Print_ocaml_compilers.cmd ]
+  [ Init.cmd; Pin.pin_cmd; Pin.unpin_cmd; Update.cmd; Pull.cmd; Opam_install.cmd;
+    Depext.cmd; Print_ocaml_compilers.cmd ]
 
 let () = Cmdliner.Term.(exit @@ eval_choice Default.cmd cmds)
--- a/cli/common.ml
+++ b/cli/common.ml
@@ -1,5 +1,4 @@
 module Ffmt = Fmt
-open Stdune
 open Duniverse_lib
 
 module Arg = struct
@@ -39,6 +38,17 @@ module Arg = struct
         windows_only priority extra_path
     in
     Cmdliner.Term.env_info ~doc var
+
+  let dev_repo =
+    let parse s =
+      match Opam.Dev_repo.from_string s with
+      | { vcs = Some Git; uri = dev_repo_uri } ->
+          (match Uri.host dev_repo_uri with
+          | Some _host -> Ok dev_repo_uri
+          | None -> Error (`Msg "dev-repo without host"))
+      | { vcs = None | Some (Other _); _ } -> Error (`Msg "dev-repo doesn't use git as a VCS") in
+    Cmdliner.Arg.conv ~docv:"DEV_REPO" (parse, Uri.pp_hum)
+
 
   let caches =
     let duniverse_cache =
@@ -90,6 +100,7 @@ end
 
 (** Filters the duniverse according to the CLI provided list of repos *)
 let filter_duniverse ~to_consider (src_deps : _ Duniverse.Deps.Source.t list) =
+  let open Stdune in
   let open Rresult in
   match to_consider with
   | None -> Ok src_deps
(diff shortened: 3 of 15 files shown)


rizo:
Attempts to address https://github.com/ocamllabs/duniverse/issues/37 and is somewhat related to https://github.com/ocamllabs/duniverse/issues/85.

rizo:
Posting my personal checklist for this, just to help with future work and discussion.

### Checklist
- [x] Make sure to delete git refs from pulled pin repos if source mode is used.
- [x] If multiple pins belong to the same repository they should be aggregated.
- [x] If an opam file for a package being pinned is found pins_dir, do not fetch the source.
- [ ] Ensure that pins in config always have deps entries in duniverse.
- [x] Implement the duniverse pin/unpin commands that update the dune-get file.
- [x] Init must recompute the git refs of the pins (even if an opam file is present?).
- [ ] Ensure that the pinned package is dune-compatible (i.e. is_dune is true).
- [ ] Consider setting the package version for pins to tag, if any.
- [ ] Consider supporting opam-repo pins.
- [ ] If the fetched source is not a dune repo, warn the user.
- [ ] Ensure that if a pinned opam file has a src.url/dev-repo it is not used.
- [ ] Make sure that deps in dune-get are updated even if the pins were not fetched.
- [ ] Do not add pin if present.
- [x] Fail when removing a non-existent pin.
- [ ] Consider using all opam files found in a pinned repo.
- [ ] Use github urls (e.g. `duniverse pin foo/bar`).
- [ ] Consider renaming `init` to `solve`.
- [ ] Read `pin-depends` field from root opam files.
- [ ] Duniverse unpin must remove the pinned opam file if present.

NathanReb:
I just gave that a try and it's a delight, thanks!

I noted a couple of things that aren't on the above list and that could make this feature easier/smoother to use:
1. [ ] running `duniverse pin x ...` fails if `x` is already pin. That means one has to run `duniverse unpin x` and then `duniverse pin x ...`. It would be nice to warn the user and ask for permission to overwrite the existing pin instead.
2. [ ] the current workflow involves running `duniverse init`, `duniverse pin` and `duniverse init` again.
It might be nice for `duniverse pin` to rerun the extra bits from `duniverse init` required to keep a consistent `dune-get`. Maybe that's what you meant by "Always update dependencies when adding/removing pins." but I wasn't sure.
3. [ ] duniverse init now uses the cache so it might make sense to add the cache related options to it as well. If we chose to implement the above suggestion, they should be added to `pin`/`unpin` instead I guess.

rizo:
Thanks @NathanReb! That's a good list, agree with everything :)

avsm:
Also tested, and happy to iterate on the remaining checklist items in trunk.

PR #95, opened 2020-06-18
Add support for a tools config stanza and subcommand

avsm:
This records the versions of tools required by the repository in dune-get, with a view to making them available from a new `duniverse tools` subcommand.
    
The `tools` subcommand sets up an environment in _ocaml/bin which has every binary that the project needs to build. This includes dune and OCaml, as well as mdx, merlin, odoc etc.
    
The feature isn't quite complete yet, as we still need to flesh out the support for what happens when opam is available, and also to add a `duniverse exec` command to set the PATH appropriately.  These will follow in a later PR.


avsm:
I'm going to merge this, and note that it'll break existing dune-get files.   https://github.com/ocamllabs/duniverse/issues/96 tracks adding that support to duniverse.

PR #102, opened 2020-06-25
Confirm if an existing pin should be replaced.

rizo:
Currently duniverse will terminate with an error when adding a pin that already exists in the config file. This PR changes this behaviour to ask the user for confirmation to replace the existing pin.

Resolves the first point in https://github.com/ocamllabs/duniverse/pull/87#issuecomment-640549805

I also renamed `to_add` and `to_remove` to `pin_name`, seemed a bit better.

### DIFF
--- a/cli/pin.ml
+++ b/cli/pin.ml
@@ -1,12 +1,12 @@
 open Duniverse_lib
 open Rresult
 
-let confirm_replace_pin name () =
-  Prompt.confirm_or_abort ~yes:false ~question:(fun l ->
+let confirm_replace_pin name ~yes () =
+  Prompt.confirm_or_abort ~yes ~question:(fun l ->
     l "A pin with name %a already exists, would you like to replace it?"
       Fmt.(styled `Yellow string) name)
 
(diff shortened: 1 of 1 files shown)


cli/pin.ml
```diff
@@ -1,8 +1,12 @@
 open Duniverse_lib
 open Rresult
 
+let confirm_replace_pin name () =
+  Prompt.confirm_or_abort ~yes:false ~question:(fun l ->
```
NathanReb, line 5: I think it might be nice to add the `yes` flag to the command and to pass it here as well.

NathanReb:
I rebased and added the `--yes` flag to the pin command but unfortunately I cannot push to your branch. Could you enable me to do so for this PR?

rizo:
Sorry for the delay and thanks for the review, @NathanReb. I added you as a collaborator, you should now have write access to the forked project, feel free to push directly to this branch.

Let me know if there's anything else I can help with.

NathanReb:
Awesome, thanks!

PR #105, opened 2020-06-25
Read pin depends

rizo:
Adds the ability to read pins from the `pin-depends` field in the root opam files during `init`.

This is IMO a more convenient way to keep track of pins as they don't have to be manually added with the `duniverse pin` command. (I currently add them in a Makefile which I don't particularly like.)

Having two sources of pins means that there might be conflicts. For example, what should happen if `dune-get` has a pin `foo` that was added manually, but there is also a root package with `foo` in `pin-depends`.

Another example is when someone runs `duniverse pin bar` without realising that there's a `pin-depends` entry for `bar` in one of the opam files.

I see `dune-get` as kind of a "lock" files, so to me, when running `duniverse init`, opam pins should overwrite what's already in `dune-get`.

Thoughts?

@avsm Submitting this as a draft for now to clarify this point first. To be honest I think this is useful even without conflict resolution. We can throw an error if there're conflicts and let users deal with them manually.

### DIFF
--- a/cli/init.ml
+++ b/cli/init.ml
@@ -44,7 +44,13 @@ let run (`Repo repo)
   let open Rresult.R.Infix in
   init_local_opam_repo opam_repo >>= fun local_opam_repo ->
   Opam_cmd.find_local_opam_packages repo >>= fun local_paths ->
-  Pins.read_from_config Fpath.(repo // Config.duniverse_file) >>= fun pins ->
+
+  Pins.read_from_config Fpath.(repo // Config.duniverse_file) >>= fun config_pins ->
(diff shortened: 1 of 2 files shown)


rizo:
Oh, and, of course, one additional complication is the fact that dune currently does not support `pin-depends` in `dune-project` files that can be used to generate the opam files.

NathanReb:
I'm closing this as it is based on duniverse. @rizo feel free to open an issue if `opam monorepo` lacks features you need!

NathanReb:
It does but that PR isn't going to help with this, please open an issue if you have a need for `pin-depends` support!

PR #107, opened 2020-07-06
Remove dependency towards Stdune

samoht:


avsm:
This is conflict central now since #112...

PR #110, opened 2020-08-10
Generate opam lockfiles

NathanReb:
This PR introduces a new command that generates a `.opam.locked` file from the local `dune-get`.

The generated opam files contains the dependencies and all the pin depends necessary to have a reproducible opam workflow.

For now the opam lockfile doesn't contain all the information the `dune-get` file does but I'm looking into this so that we can totally replace the dune-get file, using `x-` extra opam fields.
One annoying thing is that the pin-depends don't allow us to track the ref initially used to do the commit resolution i.e. the tag for regular packages or the eventually the branch for the pinned ones.
The pinned packages won't cause much trouble since we'll probably record them in an extra field somewhere, with the branch information stored so if we want to run the equivalent of `duniverse update` it's still possible.
For the regular ones we still have the version but we lose the tag we inferred. It's not much work to recompute it but it still feels like we lose a bit of information and I'm wondering if it can cause issues in some cases.

Also the opam file might contain the same information but it won't be structured in a duniverse friendly way. Again, it's not blocking but it will complexify this part of the code slightly.

Just to summarize if I got this correctly, the next steps are:
1. Add extra field so that the generated lockfile holds the same information as the dune-get
2. Add functions to parse such a lock file to the internal `Duniverse.t` representation
3. Make it so `duniverse init` generates that instead and do not use `dune-get` internally

Another point worth discussing is how should we deal with multi-packages repos. Right now there's a single lockfile with all the deps for the root packages. The idea is to use it as a strict replacement for the dune-get file so we're keeping the same convention. As we discussed on slack it might make sense to generate one lockfile per package, as long as we pin the internal deps to the local package so that it "inherits" their pin-depends but that does make the lockfiles less standalone so I'm not too sure about that. Right now all the deps are described in this lockfile, including transitive ones and I'm not sure how that should translate into multiple lock files. Anyway that's definitely not necessary for our first prototype!


NathanReb:
@avsm feel free to give this a try as the fields relevant to opam won't change and this should already give you an opam installable setup.

If you want to give it a spin the workflow would be:
1. run `duniverse init` to generate a `dune-get` file if you don't already have one
2. run `duniverse to-opam` to generate the corresponding opam lockfile
3. run `opam install ./duniverse-lock.opam.locked` to install the locked dependencies using pins

It should work from there!

NathanReb:
A quick note on what's left to do:
1. [ ] the `depends` and `pin-depends` aren't actually enough to recompute the exact internal `Duniverse.t` so we either need to store it in a separate field, either just what's required so we don't duplicate the information from `depends` and `pin-depends` or a complete representation of the source and opam deps to ease the parsing of the lockfile. The latter seems alright even though it induces quite some redundancy in the lockfile.
2. [ ] decide what to do about `depexts`. I kind of liked the idea that the lockfile would be standalone, i.e. it defines all dependencies, including transitive ones + the depexts but if those are going to be installed through opam anyway, there isn't much value in doing so I guess.
3. [ ] Make it so the opam lockfile becomes the default duniverse config file, replacing `dune-get`. I'll keep the `to-opam` command I added for testing, renaming it `migrate-dune-get` so that existing users can migrate their `dune-get` file without having to recompute the lock from scratch, might be helpful for RWO for instance.

NathanReb:
We'll keep the depext field for now to maintain the same set of features!

NathanReb:
@avsm I think this has everything we need now, you should give it a spin! I'll look into re-implementing `pull` using opam's lib in the meantime!

avsm:
Looks good. It's possibly finding a few latent bugs in opam's lockfile handling: lots of the below when running on duniverse itself:

```
[0install-solver.2.17] synchronised from git+https://github.com/0install/0install.git#4a837bd638d93905b96d073c28c644894f8d4a0b
[WARNING] Lock-file /Users/avsm/src/git/ocamllabs/duniverse/_opam/.opam-switch/sources/0install-solver/0install-solver.opam.locked doesn't exist, using
          /Users/avsm/src/git/ocamllabs/duniverse/_opam/.opam-switch/sources/0install-solver/0install-solver.opam instead.
0install-solver is now pinned to git+https://github.com/0install/0install.git#4a837bd638d93905b96d073c28c644894f8d4a0b (version 2.17)
[alcotest.1.2.1] synchronised from git+https://github.com/mirage/alcotest.git#6ca7cc04ea07f5bd5b001fed4a0313d82e4b3cc0
[WARNING] Lock-file /Users/avsm/src/git/ocamllabs/duniverse/_opam/.opam-switch/sources/alcotest/alcotest.opam.locked doesn't exist, using
          /Users/avsm/src/git/ocamllabs/duniverse/_opam/.opam-switch/sources/alcotest/alcotest.opam instead.
alcotest is now pinned to git+https://github.com/mirage/alcotest.git#6ca7cc04ea07f5bd5b001fed4a0313d82e4b3cc0 (version 1.2.1)
Processing: [astring.0.8.4+dune: git]
[astring.0.8.4+dune] synchronised from git://github.com/dune-universe/astring.git#baa4e7e2bf919293aac2c899786c2ac40757dd9b
[WARNING] Lock-file /Users/avsm/src/git/ocamllabs/duniverse/_opam/.opam-switch/sources/astring/astring.opam.locked doesn't exist, using
          /Users/avsm/src/git/ocamllabs/duniverse/_opam/.opam-switch/sources/astring/astring.opam instead.
astring is now pinned to git://github.com/dune-universe/astring.git#baa4e7e2bf919293aac2c899786c2ac40757dd9b (version 0.8.4+dune)
```

It looks like opam expects both a .opam and a .opam.locked.  Things improved when I did a `mv duniverse-lock.opam.locked duniverse.opam.locked`.  Perhaps we should just pick the root package (from dune-project (name)) rather than duniverse-lock. That would let `opam switch create . --locked` work out of the box.

More worryingly, the pin-depends handling is really slow in 2.0. I'll try that in 2.1alpha3 to see if its improved.

NathanReb:
You might have a bit of a hard time trying it with opam 2.1 as the opam lib has introduced breaking changes between now and then unfortunately.

Also do you understand why changing the name of the lockfile would fix the above warnings? Is it because if it doesn't have `.opam` corresponding to the `.opam.locked`, it expects a `.opam.locked` for all the packages?

I'll change it to use the package name when there's only one and eventually prompt for user input when there's more. I think we should really think the multi-opam scenario through as it might be really confusing. Having such different `.opam` and `.opam.locked` for the main package because it includes the dependencies for all packages might come off as a suprise to users of regular `.opam.locked` files. Even worse, what if there are non co-installable packages in the repo? I think we can't go wrong if we let the user chose in those cases.

NathanReb:
Hmm now that I think about this further not having a fixed name or our own extension makes it harder to reliably find the duniverse lock file in most commands.

Using the name from the `dune-project` slightly improves this on our side but it comes with the same issues I mentioned above.

Understanding exactly why opam goes nuts if there's no corresponding `.opam` file might actually help figuring out the best solution here. I'll take a look.

NathanReb:
Just changed it so that it uses the only package name or if there are several, the project name from the `dune-project`!

NathanReb:
Is there anything blocking here or is this good to go?

One thing I was thinking we could improve before we merge is to eventually generate the lock file from the main package opam file so that the other fields (maintainer, authors, etc...) are consistent rather than filled with dummy values. For single package repos it's definitely an improvement, for multi-opam ones I'm not so sure.What do you think?

PR #112, opened 2020-09-03
Turn duniverse into an opam plugin

NathanReb:
This turns the duniverse tool into an opam plugin. For now it's just uses the tool as it was before, stripped from some features that are already present in opam, i.e. the `depext` and `tools` sub commands and their conrresponding fields in the config.

The most important next step are:
1. [x] filter out non dune packages for the solver
2. [x] Use the local opam config, switch, etc.. rather than a hardcoded one
3. [x] warn users if they didn't add dune-universe/opam-repo to their list of repos
4. [ ] pin to tarballs when possible (eventually add a flag to always git pin)
5. [x] remove pin and unpin commands as you'll now do it through opam
6. [x] allow recursively scanning for local packages rather than just at the root


NathanReb:
CC @samoht !

samoht:
Can we merge this PR (and add improvement in subsequent PRs)?

NathanReb:
Yes I think we can merge this!

avsm (approved):
Looks good to me, great work! Could you open an issue about the "pin to tarballs" feature, and then this should be good to merge?

PR #131, opened 2020-11-23
Clean up lock file and internal representations

NathanReb:
This PR reworks a little bit how we represent the duniverse and the lockfile internally.

It also makes the lockfile closer to a regular opam file. When rebuilding the duniverse representation from the lockfile it relies on regular opam fields such as `pin-depends` along with some extra fields that are required to keep the implementation simple while not relying on opam to collect pieces of data (especially since those could have changed since the lock).


NathanReb:
The CI failure is on 4.12 and seems related to a known bug in dune so I'm ignoring it for now!

PR #142, opened 2020-12-03
Update README

NathanReb:


README.md
```diff
@@ -1,66 +1,86 @@
-# duniverse -- the spice of OCaml build life
+# opam-monorepo
 
-This is an experimental vendoring system for Dune.  Not for public consumption
-just yet.
+`opam-monorepo` is an opam plugin designed to assemble standalone dune workspaces
+with your projects and all of their dependencies, letting you build it all from scratch
+using only `dune`.
```
emillon, line 5: Not sure how technical you want to be...
```suggestion
using only `dune` and `ocaml`.
```

README.md
```diff
@@ -1,66 +1,86 @@
...
+If that's not the case the solver will report which packages don't build with dune.
 
-COMMON OPTIONS
-       --help[=FMT] (default=auto)
-           Show this help in format FMT. The value FMT must be one of `auto',
-           `pager', `groff' or `plain'. With `auto', the format is `pager` or
-           `plain' whenever the TERM env var is `dumb' or undefined.
+We maintain a [separate opam repository](https://github.com/dune-universe/opam-overlays) with dune
+ports of commonly used opam packages. If you have non-dune dependencies it is advised that you add
+this repository before running `opam monorepo lock`. You can do so by running the following command:
```
emillon, line 61: I think it's worth mentioning that opam configuration (local pins, remotes) influences the lock file generation, but `pull` does not depend on that.

README.md
```diff
@@ -1,66 +1,86 @@
...
 
-       --version
-           Show version information.
+The `lock` command takes your global and switch's opam configurations into account, meaning any
+opam repository or pins you set up will be picked up by the solver when resolving the full set of
+your project's dependencies.
 
-SEE ALSO
-       dune(1), git(1), opam(1)
+The generated lockfile is meant to be compatible with `opam` in such a way that running `opam
```
emillon, line 71: Does that work? I was under the impression that this is a design goal but that's not 100% the case yet.

README.md
```diff
@@ -1,66 +1,86 @@
...
+install . --locked` should give you the same versions you would using `opam monorepo pull` in
+a reproducible way (i.e. independantly of any change that might have happened on the upstream
```
emillon, line 73:
```suggestion
a reproducible way (i.e. independently of any change that might have happened on the upstream
```

we could also say pin-dependently

README.md
```diff
@@ -1,66 +1,86 @@
...
+opam-repository) thanks to the `pin-depends`.
+You can use that property to your advantage by allowing one to choose between a "monorepo" or
+regular opam workflow depending on the situation.
 
+## Monorepo projects
 
-```
+If you wish to use `opam-monorepo` to manage your dependencies, we suggest that you git verison the
```
emillon, line 80:
```suggestion
If you wish to use `opam-monorepo` to manage your dependencies, we suggest that you git version the
```

README.md
```diff
@@ -1,66 +1,86 @@
...
 
-## duniverse
+## Usage
 
+The basic usage for opam-monorepo is to start by running the following command from the root
+of your project:
+```
+opam monorepo lock
 ```
-NAME
```
emillon, line 20: In opam-compiler I added [some rules](https://github.com/ocaml-opam/opam-compiler/blob/master/doc/create.txt) to dump the generated docs into text files, I found that was a nice tradeoff between writing markdown docs and having people install the program before they're able to read the docs.

README.md
NathanReb: I'm fine with adding ocaml!

README.md
NathanReb: You're right it probably needs clarification!

README.md
NathanReb: It does! There might be a few corner cases where it doesn't but it definitely should work in most cases.

README.md
NathanReb: I think a proper readme will look much nicer than a manpage ever will so I'd rather maintain them separately to be honest.

NathanReb: Hmm ok I see what you mean, yeah we could eventually add that separately in a subsequent PR!

README.md
NathanReb: Just pushed a little something!

PR #147, opened 2020-12-15
Include transitive depexts in the lockfile

### LINKED ISSUE #144: Adding depexts in lock file
At the moment the lock file does not contain the transitive depexts. This means that when looking for them (such as in CI), it's necessary to pin the package recursively. Here's an example from the RWO CI:

```Dockerfile
RUN opam pin -n add rwo . --locked
RUN opam depext --update -y rwo
RUN opam pin -n remove rwo
```

This is an issue because it requires access to opam-repository when building the package. Moreover, it can require a version that is as recent as the one that was used to generate the lock file. OCaml-ci does not always have that, since it assumes that it's safe to use the oldest version that knows about the locked version of ocaml, dune and opam-monorepo itself (see ocurrent/ocaml-ci#282).

NathanReb:
Fixes #144

### DIFF
--- a/opam-monorepo.opam.locked
+++ b/opam-monorepo.opam.locked
@@ -56,6 +56,24 @@ depends: [
   "uuidm" {= "0.9.7+dune"}
   "uutf" {= "1.0.2+dune"}
 ]
+depexts: [
+  ["devel/pkgconf"] {os = "openbsd"}
+  ["pkg-config"] {os-distribution = "cygwinports"}
+  ["pkg-config"] {os-family = "debian"}
+  ["pkg-config"] {os = "macos" & os-distribution = "homebrew"}
+  ["pkgconf"] {os = "freebsd"}
+  ["pkgconf"] {os-distribution = "alpine"}
+  ["pkgconf"] {os-distribution = "arch"}
+  ["pkgconf-pkg-config"] {os-distribution = "centos" & os-version >= "8"}
+  ["pkgconf-pkg-config"] {os-distribution = "rhel" & os-version >= "8"}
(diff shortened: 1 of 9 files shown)


opam-monorepo.opam.locked
```diff
@@ -462,4 +516,4 @@ x-opam-monorepo-duniverse-dirs: [
   ]
 ]
 x-opam-monorepo-root-packages: ["opam-monorepo"]
-x-opam-monorepo-version: "0.1"
+x-opam-monorepo-version: "0.2"
```
emillon, line 519:
```suggestion
x-opam-monorepo-version: "0.1"
```

This might be a strange idea, but if we just patch the lock file version here, the CI can be green without adding support to ocaml-ci right now:
- it'll install depexts (using the "wasteful" technique)
- opam monorepo pull 0.1.0 and master understand this lock file version

NathanReb: Yeah let's do that and we'll update once this is released. We'll need to think of a better way to support new features before the release in the future though!

emillon:
Locally with ocaml-ci this exhibits the dune vendoring bug, but that's also the case on master.

emillon (approved):

emillon:
To summarize the situation, this one is ready ; CI will stay red until there's a new dune release (it's not possible to downgrade to 2.6 because 0.1.0 uses dune 2.7).

NathanReb:
Ok let's merge.

Next steps are to downgrade the vendored dune repo to 2.6 so that the dialect bug isn't triggered by the host dune on the CI anymore, we'll upgrade it again once a fix has been released in dune!

PR #158, opened 2021-02-04
Remove mention of duniverse, use opam monorepo instead.

rizo:
Just a minor fix in the error message.


NathanReb (approved):
Ouh, nice catch, thanks!

Issue #153, opened 2021-01-21
Add suport for pin-depends


bobot:
If in a project I need a particular unreleased version of a package I would like to use the following workflow:

- `$pkg.opam.template` defines the `pin-depends` variable pointing to a branch
- dune generates the `$pkg.opam`
- `opam-monorepo lock` creates a lock file which use the pinned commit

I'm not fixed on this workflow, I'm open to suggestion. Currently `opam-monorepo` doesn't take the `pin-depends` into account.

NathanReb:
I think `opam-monorepo` should respect the opam files semantic and therefore that it should indeed honor pin-depends when locking an opam file.

How exactly that should be done is to be debated. One possibility is simply to pin the corresponding deps in the current switch, which is, I believe, what opam does when installing local packages for instance. I'm not to keen on having opam monorepo modifying the user's opam config (global or per switch) though but it's unclear if it's a problem in this case. I need to look further into this. Either it should be easy enough to do it without actually pinning the packages if it happens to be the right thing to do.
Do you happen to have a personal preference or opinion on the matter?

In the meantime, if you do wish to work around that limitation, you can pin those packages yourself in your opam switch, `opam monorepo` will properly interpret those! I imagine there might even be some clever opam invocation that will pin the `pin-depends` to your switch without installing anything.

bobot:
I believe that `opam-monorepo` should not in any way modify the user configuration. But I can conceive that it could be more difficult, since it requires to download the git repository to get the opam file of the pinned version, ...

(For later of course, I think opam-monorepo should work without any opam initialized, for example if through debian someone install `dune`, `ocaml` and `opam-monorepo`. In fact perhaps `opam-monorepo` should always work with its own configuration (with additional remote as needed) which could be a copy of the current switch configuration of the user, but I'm talking when knowing very few is this affair :stuck_out_tongue_winking_eye: ).

rizo:
This is a limitation that is currently preventing my duniverse-managed project from transitioning to opam monorepo.

We are currently using a forked version of duniverse with [Read pin depends #105](https://github.com/ocamllabs/opam-monorepo/pull/105). It has been working reliably for the last 6 month. Of course adding this functionality after the opam plugin refactor isn't trivial.

The implemented workflow is described in [Add pin management to duniverse #87](https://github.com/ocamllabs/opam-monorepo/pull/87#issue-428669800) in the Notes section. In short:

1. Sources for pins are fetched into the duniverse folder because:
[...]

PR #159, opened 2021-02-04
Pin depends

rizo:
This is a fairly ad-hoc implementation of pin-depends (#153).

The process is as follows:

1. `pin-depends` entries are read from all local opam files.
2. Referenced sources are fetched with opam into `/tmp/opam-monorepo/pins/`.
3. The opam files from pins are collected with their `url_src` set to the pinned url (this is similar to what opam does on `opam pin` and is important otherwise the packages will be filtered out from duniverse).
4. The opam files from pins are added to `local_opam_files` to participate in dependency resolution.
5. Opam solve gets a list of pin packages to ensure they are not filtered out from the solution.
6. Finally, when looking up opam files to create summaries, the local opam files are used (for pins only, as the local packages are excluded anyway).

I want to spend a few days testing this and might refactor the code to split up `pull_pin_depends` into smaller functions and add some tests.

@NathanReb let me know what you think about this approach. Happy to make any changes you think are necessary.

### DIFF
--- a/cli/lock.ml
+++ b/cli/lock.ml
@@ -111,6 +111,7 @@ let root_pin_depends local_opam_files =
 
 let pull_pin_depends (pin_depends : (OpamPackage.t * OpamUrl.t) list) =
   let hashes = [] in
+  let pins_tmp_dir = Fpath.v "/tmp/opam-monorepo/pins/" in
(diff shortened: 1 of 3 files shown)


rizo:
Closing in favour of #167. Thanks again, @TheLortex!

PR #160, opened 2021-02-04
Add `--keep-clone` flag to the `pull` command.

rizo:
Allows to keep the pulled sources as git repositories. This in combination with having `duniverse` in `.gitignore` results in a nice workflow for contributing changes to vendored packages.

I named the option `--keep-clone` instead of `--trim-clone=false` to keep it as a flag. I can can the name if it's not very clear.

### DIFF
--- a/cli/common.ml
+++ b/cli/common.ml
@@ -22,6 +22,10 @@ module Arg = struct
 
   let non_empty_list_opt = Cmdliner.Term.pure (function [] -> None | l -> Some l)
 
+  let keep_clone =
+    let doc = "Keep the `.git' directory after pulling the vendored soruces." in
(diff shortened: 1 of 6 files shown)


rizo:
Rebased onto `main`. Is there any interest in merging this?

I have been using this for a while to make changes to pinned dependencies as part of duniverse.

NathanReb:
It doesn't make as much sense as it used to since we default to pulling tarballs instead of always relying on git but if you have a use for this I guess it's a good enough reason!

It would be nice to have a way to propagate the local changes back into the lockfile but I guess we still need to see how people will use this option first!

Sorry for the delay, I'll look into it.

NathanReb (approved):
I updated the changelog entry to reflect the latest update and added the PR and author reference!

Thanks for working on this!

PR #163, opened 2021-03-03
Fix lockfile path inference

### LINKED ISSUE #162: Wrong behavior when locking a single package
When locking specific packages, a different behavior exists depending on the number of packages locked:
- a single package `xxx` generates `xxx.opam.locked`
- multiple packages generates `<project name>.opam.locked`

In the first case, a subsequent `opam monorepo pull` would fail as it looks for `<project name>.opam.locked` to perform the pull.

NathanReb:
Fixes #162 

The lockfile path inference mechanism was broken because it depended on which packages were locked on one hand (when running `lock`) and only on which packages were defined at the root of the project on the other (when running `pull`).

It is now symetrical and only relies on the project name as defined in the `dune-project` file so `lock` and `pull` will get to the same conclusion regarding its location. This can be considered a breaking change as single package repositories could get away without defining a name in the dune-project before and can't do that anymore. I think it is fine to ask of our users that they fill in this field.

I also added a CLI option to explicitly set the path to the lockfile, providing a bit of flexibility for specific use cases.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,9 @@
   provided, use the value of the version field in the opam files instead of the
   default `zdev` if it is defined (#183, @emillon)
 
+- Add a `-l`/`--lockfile` command line option to explicitly set the lockfile
+  to use or generate in `pull` or `lock` (#163, @NathanReb)
+
 ### Changed
 
 ### Deprecated
@@ -21,6 +24,11 @@
 - Simply warn instead of exiting when the dune-project file can't be parsed
   by `pull` as it only use it to suggest updating the lang version for
   convenience (#191, @NathanReb)
+
+- Fix a bug where `pull` and `lock` would expect the lockfile to sit in a
+  different place and pull would fail. `pull` now simply looks for a
+  `.opam.locked` file and pulls it unless there are multiple matching files in
+  the repository's root. (#163, @NathanReb)
 
 ### Removed
 
--- a/cli/common.ml
+++ b/cli/common.ml
@@ -14,6 +14,16 @@ module Arg = struct
         & opt fpath (Fpath.v (Sys.getcwd ()))
         & info [ "r"; "repo" ] ~docv:"TARGET_REPO" ~doc)
 
+  let lockfile =
+    let doc =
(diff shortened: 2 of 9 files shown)


TheLortex:
Hello, it would be nice to have this merged ! 
For the breaking change, a solution would be to keep having a permissive lock step (= without requiring a name in the dune-project file), and adapt the pull step to pull if there is a single `*.opam.locked` file in the folder, and fail otherwise (requiring to use the `-l` option or a name in the dune project file). What do you think ?

pitag-ha (commented):
Leaving aside the breaking change behavior @TheLortex has mentioned, LGTM.

pitag-ha:
> if there is a single *.opam.locked file in the folder

Out of curiosity: in which case(s) are there several lock files in the work space? Might it happen with that solution that there is exactly one lock file in the work space that's not the one that should be used for pulling, so `pull` would succeed (pulling the wrong packages) even if `lock` failed or hasn't been run?

TheLortex:
@pitag-ha When a project defines multiple packages, it may happen that the user performs locks of specific packages. In that case different lockfiles could be generated.

avsm:
Another usecase of multiple lockfiles is for different OCaml compiler versions.

NathanReb:
I'll revive this PR, sorry for the delay!

Having thought about and discussed this topic with @TheLortex. I think `pull` doesn't need to look for the lockfile in a specific place. It could instead look for opam-monorepo lockfiles in the current directory. If there's only one, it pulls it, otherwise it fails and tells the user they have to explicitly specify the path through the command line.

NathanReb:
With that in mind, I don't think the changes on `lock` are necessary anymore. I think this PR should stick to:
- Adding the `--lockfile` option to both `lock` and `pull`, it makes sense that this is configurable
- Modifying `pull` so that it looks for all lockfiles at the repo's root, screams when there's none, pulls it when there's only one and ask for help when there's more than one.

For that second part, to avoid breakage we could make it so that if there are several lockfiles it falls back to using the one named after the dune project but I don't think it's worth doing so as I don't expect a lot of our users to have several lockfiles anyway.

emillon (approved):

PR #164, opened 2021-03-08
Exclude packages depending on jbuilder by default

TheLortex:
Since dune 2.0, `jbuild` files are not supported. As a consequence, if `opam monorepo` was to vendor a project depending on `jbuilder`, it would break the whole monorepo build.   

There are two changes in this PR:
- by default: don't include packages depending on `jbuilder` for the lockfile resolution.
- add a `--allow-jbuilder` option to enable the old behavior.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


NathanReb (approved):
Awesome, thanks!

PR #166, opened 2021-05-05
Fix git default branch computation

TheLortex:
When given an opam git remote in the form `git+https://github.com/...`, `git ls-remote --symref <remote> HEAD` would fail as it doesn't recognize urls in the form `git+https`, because this is something specific to opam. 

This PR applies the same mechanism as in `resolve_ref`, which is to remove the `git+` prefix in that situation.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,7 @@
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


NathanReb (approved):
Looks great, thanks!

PR #167, opened 2021-05-27
Pin depends

TheLortex:
This is a rebase of @rizo's PR: https://github.com/ocamllabs/opam-monorepo/pull/159

Fixes https://github.com/ocamllabs/opam-monorepo/issues/153

### DIFF
--- a/cli/lock.ml
+++ b/cli/lock.ml
@@ -180,9 +180,6 @@ let pull_pin_depends ~global_state
   if List.is_empty pin_depends then Ok OpamPackage.Name.Map.empty
   else
     let pins_tmp_dir = Fpath.v "/tmp/opam-monorepo/pins/" in
-    let cache_dir =
-      OpamRepositoryPath.download_cache global_state.OpamStateTypes.root
-    in
     Logs.debug (fun l ->
         l "Pulling pin depends: %a"
           Fmt.(list ~sep:(unit " ") Fmt.(styled `Yellow string))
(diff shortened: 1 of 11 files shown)


cli/lock.ml
```diff
@@ -109,6 +109,43 @@ let root_depexts local_opam_files =
...
+let root_pin_depends local_opam_files =
+  OpamPackage.Name.Map.fold
+    (fun _pkg (_version, opam_file) acc -> OpamFile.OPAM.pin_depends opam_file @ acc)
+    local_opam_files []
+
+let pull_pin_depends (pin_depends : (OpamPackage.t * OpamUrl.t) list) =
+  (* TODO: Group pin-depends with the same url. *)
+  if List.is_empty pin_depends then Ok OpamPackage.Name.Map.empty
+  else
+    let pins_tmp_dir = Fpath.v "/tmp/opam-monorepo/pins/" in
```
emillon, line 121: Can we use a tempdir function like `Bos.OS.Dir.with_tmp` here?

cli/lock.ml
```diff
@@ -109,6 +109,43 @@ let root_depexts local_opam_files =
     (fun _pkg (_version, opam_file) acc -> OpamFile.OPAM.depexts opam_file :: acc)
     local_opam_files []
 
+let root_pin_depends local_opam_files =
+  OpamPackage.Name.Map.fold
+    (fun _pkg (_version, opam_file) acc -> OpamFile.OPAM.pin_depends opam_file @ acc)
+    local_opam_files []
+
+let pull_pin_depends (pin_depends : (OpamPackage.t * OpamUrl.t) list) =
+  (* TODO: Group pin-depends with the same url. *)
```
emillon, line 118: Can this be taken care of or a ticket created for this?

emillon:
I just rebased and added a changelog entry.

NathanReb:
Thanks for the multiple rebase, I'll revive this PR.

There are a couple things that I'd like to improve, in particular the interface for `Opam_solve` and code sharing for pull and pulling the pin depends.

cli/lock.ml
NathanReb: I implemented this bit!

cli/lock.ml
NathanReb: I implemented it and removed the comment!

NathanReb:
I made a few improvements:
- Clean up the interface of `Opam_solve` and let it deal with pin_depends rather than merging them into `local_opam_files` in `Lock`. `calculate` now expects a map for local packages and a map for pin-depends, abstracting everything else away from `Lock`.
- I moved the pin_depends extraction further down the chain so that we only load opam states once
- I fixed a small quirk in `Opam_solve.get_opam_info` so that the pin-depends map is only looked up once instead of twice (`find_opt` instead of `mem` + `find`)
- I forbid having local packages in a pin-depends field. Before `lock` would have allowed it but the pin would effectively be ignored since local packages are removed from the solution in the lockfile. There's no way to properly support this as you either want it local and use whatever you have in your repo or use a pin that will be pulled in your duniverse. Having both will lead to a dune error because you have the same package twice in your workspace.
- I factored opam pulling code for pin-depends and the pull command
- I added deduplication of the repository's pin-depends (the same pin could appear several times in different opam files)
- I added rejection of different pins for the same package
- we group pin-depends per URL, meaning if several packages are pinned to the same URLs, the repo is pulled once instead of once per package
- I implemented @emillon suggestion to use `Bos` to generate a temporary directory

I need to test this a bit.
@TheLortex would you mind trying this out for mirage and see if it still works?
@emillon could you give this another round of review?

emillon (approved):

PR #178, opened 2021-06-07
Discard ocaml variants by default

NathanReb:
This was causing the solver to select unreleased ocaml versions sometimes because their dependency could be met using packages such as `ocaml-variants.4.11.3+trunk`.

The variant selected this way made no particular sense as well since the solver was picking the highest version number according to the `+<variant-feature>` suffixes.

The solution is now required to include an `ocaml-base-compiler` unless one the local packages explicitly depends on an `ocaml-variants` package or one of the `ocaml-option-*` packages.

### DIFF
--- a/lib/opam.ml
+++ b/lib/opam.ml
@@ -1,13 +1,17 @@
 open Import
 
-let depends_on_dune ~allow_jbuilder (formula : OpamTypes.filtered_formula) =
-  let dune = OpamPackage.Name.of_string "dune" in
-  let jbuilder = OpamPackage.Name.of_string "jbuilder" in
-  let is_duneish name =
-    let eq n n' = OpamPackage.Name.compare n n' = 0 in
-    eq dune name || (allow_jbuilder && eq jbuilder name)
+let depends_on_any packages (formula : OpamTypes.filtered_formula) =
+  let packages = List.map ~f:OpamPackage.Name.of_string packages in
+  let is_one_of_packages name =
+    List.exists packages ~f:(fun p -> OpamPackage.Name.compare p name = 0)
   in
-  OpamFormula.fold_left (fun acc (name, _) -> acc || is_duneish name) false formula
+  OpamFormula.fold_left (fun acc (name, _) -> acc || is_one_of_packages name) false formula
+
+let depends_on_compiler_variants formula = depends_on_any [ "ocaml-variants" ] formula
+
+let depends_on_dune ~allow_jbuilder (formula : OpamTypes.filtered_formula) =
+  let packages = if allow_jbuilder then [ "dune"; "jbuilder" ] else [ "dune" ] in
+  depends_on_any packages formula
 
 let pull_tree ~url ~hashes ~dir global_state =
   let dir_str = Fpath.to_string dir in
--- a/lib/opam.ml
+++ b/lib/opam.ml
@@ -7,7 +7,33 @@ let depends_on_any packages (formula : OpamTypes.filtered_formula) =
   in
   OpamFormula.fold_left (fun acc (name, _) -> acc || is_one_of_packages name) false formula
 
-let depends_on_compiler_variants formula = depends_on_any [ "ocaml-variants" ] formula
+let ocaml_options =
+  (* ocaml-options-vanilla is purposefully excluded from that list as it doesn't imply
+     a dependency towards ocaml-variants on its own. *)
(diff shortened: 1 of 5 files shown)


lib/opam.ml
```diff
@@ -1,13 +1,43 @@
...
+    List.exists packages ~f:(fun p -> OpamPackage.Name.compare p name = 0)
   in
-  OpamFormula.fold_left (fun acc (name, _) -> acc || is_duneish name) false formula
+  OpamFormula.fold_left (fun acc (name, _) -> acc || is_one_of_packages name) false formula
+
+let ocaml_options =
+  (* ocaml-options-vanilla is purposefully excluded from that list as it doesn't imply
+     a dependency towards ocaml-variants on its own. *)
+  [
+    "ocaml-option-32bit";
```
emillon, line 14: FYI, if you want to be future proof you can probably generate this list using `Ocaml_version`: these packages exist only since `Since.options_packages` and they are listed in `Opam.V2.additional_packages`.

dra27:
Instead of blocking the whole of ocaml-variants, why not simply filter out the packages marked `avoid-version` (which all the trunk packages are). I assume that the solver that's running is opam 2.1, hence it's not requiring the `ocaml-beta` package (which would normally make these unavailable).

dra27:
`avoid-version` in the main solvers adds a lag to the LP problem which causes them only to be selected if there's absolutely no other way. For the `0install` solver (which I believe is what you're using?) you could achieve the same with a double-solve - i.e. try to solve with the `avoid-version` ocaml packages not there and retry if that fails (if unreleased versions actually should be being selected)

NathanReb:
I'm not sure I understand what's the problem with the current solution compared to the double solve you're proposing. Do you have a concrete example of a `depends` field content that can't be solved without selecting a variant but that would have a good reason not to explicitly specify that a variant must be selected?

dra27:
The current solution deletes an entire package in order to eliminate some of its versions. It seems a sledgehammer when the actual packages you won't to omit are actually marked?

dra27:
The double-solve is just how you _could_ implement `avoid-version` using the `0install` solver - it doesn't have the ability to express the lag which is used in the LP solvers (I have no idea whether that's useful, it was just an observation!)

avsm:
I think the problem here is that `opam-version` field isn't set at all, and so the `ocaml-beta` constraint isn't kicking in due to the `opam-version < 2.1` constraint being false (since opam-version is NULL).  If you set the `opam-version` field to the solver then this whole PR may not be necessary. @TheLortex ran into something similar a few days ago with the docs-ci-solver.

TheLortex:
I solved this issue in docs-ci by overriding the `env` passed to the solver: https://github.com/ocurrent/ocaml-docs-ci/commit/5d7f42eef9f0b472b378bd1069e47e613b674196

NathanReb:
I might be wrong as I'm not super familiar with the behaviour of that `opam-version` field regarding solving but I'm not sure that's relevant for opam-monorepo. We use a `Switch_context` and not a `Dir_context` as in the code linked by @TheLortex , which doesn't offer that kind of configuration. I would expect it's doing the right thing on its own here. If not, we should probably open an issue on opam-0install-solver so that it either is configurable there as well or that it extracts that information from the switch itself if possible.

My impression is that it already gets that information from opam itself and that this doesn't apply here.

lib/opam.ml
NathanReb: I looked into this but it unfortunately doesn't do what you suggest:
```ocaml
let additional_packages t =
      if Has.options_packages t then
        match Configure_options.of_t t with
        | Ok []
        | Error _ -> []
        | Ok options ->
            let options_only_package =
              List.map Configure_options.to_string options
              |> String.concat "-"
              |> (^) "ocaml-options-only-" in
            [options_only_package]
      else
        []
```

IIUC, It returns the `ocaml-option-*` package corresponding to the version, e.g. for `4.12.0+flambda+afl` it will return `ocaml-option-only-flambda-afl`.
I'd be happy to externalize the maintenance of this all logic as much as possible, eventually not even having to deal with it in opam-monorepo at all and have the solver do the right thing according to the user's dependencies spec but in the meantime we need a fix for it.

avsm:
It might be down to the version of opam-libs you link against, which will set `opam-version` to 2.1.

PR #180, opened 2021-07-07
Filter all local packages out for the lockfile

pitag-ha:
Before, only the packages for which the dependencies are calculated were excluded from the lockfile. To avoid build errors due to duplicated packages, all packages in the workspace need to be excluded.


pitag-ha:
I also have a version that filters the packages that are in the workspace but not in the "project" (in the sense that you take them into account to compute dependencies) from the `pin-depends`, but not from the `depends`. But from our last conversation, @NathanReb , I think what we want is excluding them totally from the lockfile, right?

lib/repo.ml
```diff
@@ -35,6 +35,17 @@ let local_packages ~recurse ?filter t =
                 "Conflicting definitions for the same package have been found:\n- %s\n- %s"
                 (Fpath.to_string a) (Fpath.to_string b)))
 
+let all_local_packages_names repo =
+  let open Result.O in
+  Bos.OS.Dir.exists repo >>= fun exists ->
+  if not exists then Ok OpamPackage.Name.Set.empty
```
pitag-ha, line 41: This is to be consistent with `local_packages`. Should that be an error instead?

lib/repo.ml
```diff
@@ -35,6 +35,17 @@ let local_packages ~recurse ?filter t =
...
+  else
+    Bos.OS.Path.fold
+      ~elements:(`Sat (fun p -> Ok (Fpath.has_ext ".opam" p)))
+      (fun path acc ->
+        OpamPackage.Name.Set.add (OpamPackage.Name.of_string Fpath.(basename (rem_ext path))) acc)
+      OpamPackage.Name.Set.empty [ repo ]
```
pitag-ha, line 47: We could also re-use `local_packages`, but I figured directly turning the packages into a set might be simpler and more efficient.

pitag-ha:
Should I add some test?

PR #183, opened 2021-07-22
Use version field of local package

emillon:
If a root package has a "version" field in its opam file, use that version instead of "zdev". This allows breaking some circular test-dependencies.

Example use case: suppose we are building a monorepo with core and its test-dependencies. core.v0.14.1 uses async in its tests (this is not expressed in the public opam metadata), but async depends on core.
Adding `"async" {with-test}` fails in the latest release of opam-monorepo because async requires core but no async version is compatible with `core.zdev`.
On the other hand, if opam-monorepo knows that the root core package is at version v0.14.1, it can (correctly) select async.v0.14.0.


emillon:
Let me know what you think. As a cleanup step I suggest to rename `Types` to `Opam` or something like that and adding an mli file.

NathanReb (commented):
The feature is definitely a great addition!

The `Types` module is legacy and we've been trying to get rid of it for a while now. The current approach is to stop using the types there and migrate to opam libraries types when writing new code, as much as possible.

Here it's a bit tricky but I'd suggest you at least move new functions and values to the already existing `Opam` module. Same goes for the test, I'd much rather not create a test file for `Types`.

A good follow up here would be to change the `package` CLI arg to return an `OpamPackage.t` instead. I can take care of this once this PR is merged!

emillon:
I just moved that to `Opam` and moved `Types.Opam` to `Opam.Package` using better types (`OpamPackage.Name.t` and `OpamPackage.Version.t` instead of strings). `OpamPackage.t` cannot be used directly because it assumes a version number. Some interfaces (e.g. in `Repo`) still use strings to represent package names but it's a bit more involved to switch that to `OpamPackage.Name.Map.t` since the map interfaces are different. I can split that move into a different PR if you prefer.

emillon:
One thing I noticed is that most of this `Opam.Package` module (or `Types.Opam`) is used to manipulate (name, version option) pairs, but the versions are only useful when an explicit version is passed on the command line, e.g. `opam monorepo lock pkg1.ver1 pkg2.ver2`. I don't know about that feature, but if it is not necessary that whole module can go away and `OpamPackage.Name.t` be used instead. If this is indeed supported I think this should be reflected in the manpage.

NathanReb:
I think we should get rid of that module entirely and use `OpamPackage.t` since all packages are ultimately assigned a version (the default zdev).

Let's do that in a separate PR.

Passing an explicit version is indeed supported and is useful to solve the same kind of problems than this PR in a context where the version field isn't set. You're right that it should be documented!

emillon:
All right. I left the `Opam.Package` parts out, so this PR should be ready. As followups I'll address:
- missing documentation for `LOCAL_PACKAGE` syntax
- using `OpamPackage.t` instead of `Types.Opam.package` (and name in the CLI)

NathanReb (approved):
Awesome possum!

PR #184, opened 2021-07-28
Document LOCAL_PACKAGE syntax for lock

emillon:
These can have version numbers.

### DIFF
--- a/cli/lock.ml
+++ b/cli/lock.ml
@@ -240,8 +240,11 @@ let allow_jbuilder =
 
 let packages =
   let doc =
-    "Explicit list of local packages to compute the lockfile from. When none \
-     are provided, all packages that have an opam file at the root of the \
+    "Explicit list of local packages to compute the lockfile from. These can \
+     be either plain package names (such as \"dune-release\"), or packages \
+     with a version number (e.g. \"irmin.2.7.1\"). Version numbers are used to \
+     tell the solver what is the version of the local package. When none are \
+     provided, all packages that have an opam file at the root of the \
      repository are used."
   in
   let docv = "LOCAL_PACKAGE" in


cli/lock.ml
```diff
@@ -159,8 +159,10 @@ let allow_jbuilder =
 
 let packages =
   let doc =
-    "Explicit list of local packages to compute the lockfile from. When none are provided, all \
-     packages that have an opam file at the root of the repository are used."
+    "Explicit list of local packages to compute the lockfile from. These can be either plain \
+     package names (such as \"dune-release\"), or packages with a version number (e.g. \
+     \"irmin.2.7.1\"). When none are provided, all packages that have an opam file at the root of \
```
NathanReb, line 164: It's probably worth mentioning what the version number are used for here otherwise users might get confused. They could for instance imagine that it will lock using whatever version you provided from opam, whereas it will simply use the local package, assigning it the provided version.

How about something along those lines: "When no version is provided, the packages are assigned the `zdev` version before querying the solver. Assigning a version can help in cases where local packages are part of the dependency cone and the `zdev` version is rejected because it doesn't fit the constraints, typically an upper bound."

It might require a bit of reordering to avoid confusion.

NathanReb (commented):
It looks good!

I left one comment with a suggestion for a slightly more detailed documentation about those explicit versions.

Let me know what you think!

cli/lock.ml
emillon: Hmm ok I see. With that in mind, and now that we read version numbers from opam files, is that feature still useful? Users can set a version field in the opam file instead. This is a bit easier to do again since it does not rely on remembering the exact `opam monorepo pull ...` command (the source of data is moved from the arguments to the files). I would think that users of opam-monorepo are in a position to edit the opam file. What do you think?

NathanReb: There are a few reasons I can think of that make me want to keep that feature (numbers or for reference, not order of significance):
1. more often than not, the opam files have no version field so they will have to add them in the first place
2. setting the version is useful for the solver and getting around solver failures, this can be an iterative process and having to edit a bunch of different files that all look very much alike from the inside isn't a great workflow here
3. in some contexts, you don't want to edit such files because you want to send patches upstreams and that is something you'll have to filter out
4. when using opam-monorepo in scripts, it's much more straightforward to pass different arguments to `lock` than to edit files which do not always use the same naming convention (`opam` vs `package.opam`), are spread in a bunch of different places etc...
5. Adding such a field has an impact on more than opam-monorepo's solver which you might want to avoid in some cases

We already use that feature to assemble the ppx-universe. I agree that if one wants a reproducible `lock` step they will use version fields but I think this feature is useful for different reasons, in different context, in particular in a dev workflow when you're trying to figure things out.

Does that sound good enough to keep it around for you?

emillon: Yes, sounds good. There's some overlap so I'm making sure we're not missing anything.

emillon:
I rebased and added a sentence about that.

NathanReb (approved):

PR #186, opened 2021-07-28
Unversion duniverse

NathanReb:
The `duniverse/` folder is now ignored and one has to run `opam monorepo pull` to stay in sync.

It will make reviews and the history much cleaner.

I also added a CONTRIBUTING.md to provide a bit of guidance on how to setup one's developement environment.

One consequence of this is that we have to use a lockfile format compatible with the latest released version but this is a good thing and was pretty much already required to keep the CI going.


NathanReb:
It does break the other builds, I'll look into it!

emillon:
OK, what's happening on windows is:

- GHA installs a git version that does newline translation: after checkout, the dune-project file will be CRLF terminated.
- `Bos.OS.File.read_lines` splits just on LF (it's a combination of `open_in_bin`, which disables in-channel newline translation, and `input_line`, which splits on just LF)
- this means that lines have a trailing CR character
- `Dune_file.is_stanza` just checks for prefix, so returns true for `(lang dune 2.6)\r`
- `Dune_file.parse_stanza` checks that `)` is a suffix, so it fails

There are a couple ways to fix that:
- I think we should support CRLF terminated dune files, so reconfiguring git or using `dos2unix` is not a solution
- we could add a `read_lines` function that opens a file in text mode and uses `input_line`. it would do the right thing
- we could insert a call to a whitespace-trimming function in `parse_stanza`, such as `Astring.String.trim`.

NathanReb:
> I think we should support CRLF terminated dune files
Definitely agree there. I think the best solution would be to write our own CRLF compatible `read_lines`. We should probably send a patch to `Bos` as well if we want to keep using it but I'm not sure we should.
I'm guessing that `base` and `core` are much more compatible with windows aren't they? Might be good to consider using them instead.

emillon (approved):

PR #187, opened 2021-07-28
Move Types into Lock.Package_argument

emillon:
This gets rid of the `Types.Opam` module. This abstraction (name with an optional version) is only a CLI detail, so move it into the CLI part. The rest uses opam types (or strings as previously).


NathanReb (approved):
Looks good, thanks!

PR #188, opened 2021-07-28
Make lock 2x faster by loading repo state once

emillon:
If no `~rt` argument is passed to `OpamSwitchState.with_`, it will load the repository state itself. We already load this state in `check_repo_config`, so this was causing it to be loaded twice.

Instead, we can load it manually and pass it as argument and it will use it.

Since that function dominates lock time, this makes lock twice as fast (on opam-monorepo itself: 12.4s down to 6.7s).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,9 @@
 
 ### Fixed
 
+- Improve `lock` performance (about 2x faster) by loading the repository state
+  only once (#188, @emillon)
+
 ### Removed
 
 ### Security
--- a/cli/lock.ml
+++ b/cli/lock.ml
@@ -86,31 +86,29 @@ let is_duniverse_repo (repo : OpamTypes.repository) =
   let url = OpamUrl.to_string repo.repo_url in
   String.equal url Config.duniverse_opam_repo
 
-let check_repo_config ~global_state ~switch_state =
-  OpamRepositoryState.with_ `Lock_none global_state (fun repo_state ->
-      let repos = current_repos ~repo_state ~switch_state in
-      let dune_universe_is_configured =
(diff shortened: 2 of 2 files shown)


NathanReb (approved):
This is both good and bad news, I was expecting we could spare the cost of loading the repository state if the solver succeed in finding a solution, it appears we will always have to run this costly operation.

I guess that will provide an insentive to improve the situation with state loading in opam as the performances currently are quite a disaster.

Out of curiosity, do you know if the repo_state can be extracted from the switch state directly by any chance? I.e. if we can obtain it without calling `OpamRepositoryState.with_`.

NathanReb:
Hmm what happened to the changelog check btw?

NathanReb:
Can you add a changelog entry for this?

emillon:
> Out of curiosity, do you know if the repo_state can be extracted from the switch state directly by any chance? I.e. if we can obtain it without calling `OpamRepositoryState.with_`.

I haven't found a way to do that as a shortcut, but in terms of performance it would be the same (I agree that keeping the implicit loading would read a bit better).

OTOH I think that we can expose a way to list the repositories without loading a switch state, but that state is used later to call the solver.

NathanReb:
> OTOH I think that we can expose a way to list the repositories without loading a switch state, but that state is used later to call the solver.

I think that as you're suggesting, I don't think there's any value in that for us. Furthermore, I assume a switch state is needed because we do not want a bare repository list but the specific list of repos enabled for the current switch so to me, it make sense that we're getting it from the SwitchState.

My impression is that those states contain a lot more information than we need most of the time and that lazily loading things and exposing the proper functions to access the proper subset of information without triggering a full load is what we'd need the most from opam libraries.

PR #191, opened 2021-07-29
Fix dune-project handling under windows

NathanReb:
This used to crash under some circumstances on windows machines because it wouldn't handle CRLF properly.
The new code uses regexp to properly extract the version without being sensitive to leftover CR and replace just the version number, without modifying the rest of the file, including of course, CRLF.

I also updated the code that appends a lang stanza even though it's unlikely we'll need it again for a very long time. It now tries to
respect the newline convention found inside the file, defaulting to good old `\n` if it can't figure it out.

Finally, I added a bunch of unit tests for this in the process.

### DIFF
--- a/lib/dune_file.ml
+++ b/lib/dune_file.ml
@@ -35,8 +35,7 @@ module Lang = struct
     match (int_of_string_opt major_str, int_of_string_opt minor_str) with
     | Some major, Some minor -> Ok (major, minor)
     | _ ->
-        Printf.ksprintf
-          (fun msg -> Error (`Msg msg))
+        Rresult.R.error_msgf
           "Invalid dune-project file: invalid lang version %s.%s" major_str
           minor_str
 
@@ -46,8 +45,7 @@ module Lang = struct
     | _ ->
         (* We match on [bos], meaning the only possible case here is the empty
            list *)
-        Printf.ksprintf
-          (fun msg -> Error (`Msg msg))
+        Rresult.R.error_msg
           "Invalid dune-project file: It does not start with a valid lang \
            stanza"
 
--- a/lib/dune_file.ml
+++ b/lib/dune_file.ml
@@ -3,39 +3,77 @@ open Import
 module Lang = struct
   type version = int * int
 
+  let pp_version fmt (major, minor) = Format.fprintf fmt "%d.%d" major minor
+
+  let version_to_string version = Format.asprintf "%a" pp_version version
+
+  let stanza_regexp =
+    let open Re in
+    seq
+      [
+        bol;
+        char '(';
+        rep blank;
+        str "lang";
+        rep1 blank;
+        str "dune";
+        rep1 blank;
+        group (rep1 digit);
+        char '.';
+        group (rep1 digit);
+        rep blank;
+        char ')';
+        rep blank;
+        opt (char '\r');
+        eol;
+      ]
+
+  let from_match group =
+    let major_str = Re.Group.get group 1 in
+    let minor_str = Re.Group.get group 2 in
+    match (int_of_string_opt major_str, int_of_string_opt minor_str) with
+    | Some major, Some minor -> Ok (major, minor)
+    | _ ->
+        Printf.ksprintf
+          (fun msg -> Error (`Msg msg))
+          "Invalid dune lang version: %s.%s" major_str minor_str
+
+  let from_content content =
+    let open Result.O in
+    let cregexp = Re.compile stanza_regexp in
+    match Re.all cregexp content with
+    | [] -> Ok None
+    | [ group ] -> from_match group >>= fun version -> Ok (Some version)
+    | _ ->
+        Printf.ksprintf
+          (fun msg -> Error (`Msg msg))
+          "Invalid dune-project file: Multiple lang stanzas"
+
+  let update_stanza ~version group =
+    let new_ver = version_to_string version in
+    let old_ver =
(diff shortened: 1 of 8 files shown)


lib/dune_file.ml
```diff
@@ -3,39 +3,77 @@ open Import
...
+    match (int_of_string_opt major_str, int_of_string_opt minor_str) with
+    | Some major, Some minor -> Ok (major, minor)
+    | _ ->
+        Printf.ksprintf
+          (fun msg -> Error (`Msg msg))
+          "Invalid dune lang version: %s.%s" major_str minor_str
+
+  let from_content content =
+    let open Result.O in
+    let cregexp = Re.compile stanza_regexp in
```
emillon, line 43: `compile` is the "expensive" operation, so it makes more sense to do it in the toplevel definition above rather than each time.

lib/dune_file.ml
```diff
@@ -3,39 +3,77 @@ open Import
...
+    let stanza = Re.Group.get group 0 in
+    Base.String.substr_replace_first ~pattern:old_ver ~with_:new_ver stanza
+
+  let update ~version content =
+    let cregexp = Re.compile stanza_regexp in
+    Re.replace ~all:false cregexp ~f:(update_stanza ~version) content
 
   let stanza version = Format.asprintf "(lang dune %a)" pp_version version
+
+  let append ~version content =
```
emillon, line 66: I think this should be called `prepend` since we're adding the stanza at the beginning

lib/dune_file.ml
```diff
@@ -3,39 +3,77 @@ open Import
...
-    | Some [ "lang"; "dune"; version ] -> parse_version version
-    | _ -> Error (`Msg (Printf.sprintf "Invalid lang stanza: %s" s))
+  let version_to_string version = Format.asprintf "%a" pp_version version
 
-  let is_stanza s = String.is_prefix ~prefix:"(lang " s
+  let stanza_regexp =
+    let open Re in
+    seq
+      [
+        bol;
```
emillon, line 14: the lang stanza is only valid as first line, so if you want you can change that to `bos` and remove the error case corresponding to multiple stanzas below

lib/dune_file.ml
NathanReb: I thought that once compiled, it has some kind of state and that you can't really re-use it. What is compile doing then?

emillon: No, that's the other way around: `compile` is turning an AST of regexp into an automaton using some graph science, maybe doing some optimizations on the way. You can find an example in the ocaml-re benchmarks where the regexp is compiled once and executed on many inputs:

https://github.com/ocaml/ocaml-re/blob/42c84b0f39117ece63334995258487db0878653d/benchmarks/benchmark.ml#L107-L111

lib/dune_file.ml
NathanReb: Can you confirm this has always been the case?

lib/dune_file.ml
NathanReb: You're right, `List.append` confused me here.

lib/dune_file.ml
NathanReb: One thing worth noting is that if we do this, we won't be able to detect bad dune-projects anymore.

I think we can safely only accept them, and simply warn that the dune-project is invalid if we can't find the lang stanza (instead of prepending it).

NathanReb:
I applied some of your comments. The last one, about matching agains `bos` as only a `dune-project` starting with a `lang` stanza on the first line is considered valid, lead me to change the rest of the behaviour a bit. We now only allow for such valid dune-project files, if it's not, we simply output a warning and proceed with pulling as this step is here for convenience and is not strictly required!

lib/dune_file.ml
```diff
@@ -1,41 +1,73 @@
...
+           eol;
+         ])
+
+  let from_match group =
+    let major_str = Re.Group.get group 1 in
+    let minor_str = Re.Group.get group 2 in
+    match (int_of_string_opt major_str, int_of_string_opt minor_str) with
+    | Some major, Some minor -> Ok (major, minor)
+    | _ ->
+        Printf.ksprintf
```
emillon, line 38: This is `Rresult.R.error_msgf "..." major_str minor_str`

lib/dune_file.ml
```diff
@@ -1,41 +1,73 @@
...
+          "Invalid dune-project file: invalid lang version %s.%s" major_str
+          minor_str
+
+  let from_content content =
+    match Re.all stanza_regexp content with
+    | [ group ] -> from_match group
+    | _ ->
+        (* We match on [bos], meaning the only possible case here is the empty
+           list *)
+        Printf.ksprintf
```
emillon, line 49: same here, or even `Rresult.R.error_msg "..."`

emillon (approved):
Thanks! 2 minor suggestions and this can go!

lib/dune_file.ml
emillon: > Can you confirm this has always been the case?

I think that this forward compatibility has been considered since the release of dune 1.0.0.

NathanReb:
I kind of wanted to migrate away from `Rresult` on purpose but you're right, better to do it all at once later.

NathanReb:
It looks very much like the github actions are broken. They don't seem to be adding a lot of value to me but they sure cause a bit of troubles so I suggest we simply drop them for now until we have something reliable to build on windows and MACos.

As you already pointed out, simply building on those platforms isn't really a problem, running the actual program is.

PR #192, opened 2021-08-02
Access repo state through switch state

emillon:
This simplifies a bit how states are loaded:
- repo state is automatically loaded when switch state is loaded
- repo state can be accessed as a field of switch state


emillon:
When checking if there's such a function in #188 I forgot to check the type, it's exposed and it's a record type so we can simplify this.

NathanReb (approved):
Looks good, thanks!

PR #194, opened 2021-08-06
Make OpamSwitchState load files lazily (makes lock 4x faster)

emillon:
Profiling `opam monorepo lock` reveals that it is spending a lot of time parsing
files (in particular, lexing them) and in `OpamPp` combinators.
We cannot do a lot about these, but this dominates the profile because opam
parses so many files, tens of thousands of them. It turns out that it parses all
files from all the repositories it knows about.

This PR changes the internals of `OpamRepositoryState` and `OpamSwitchState` so
that opam files are parsed only lazily. The name and version of each package is
already known just with the file path, and the contents are often not necessary
at all so they can be hidden behind a `lazy`.

Some code is left stubbed out (it does not seem to matter to opam-monorepo at
least), but the performance is better: a lock on opam-monorepo itself now takes
1.5s (from 6.5s on `main`).


emillon:
I couldn't reproduce this after upgrading to opam 2.1 but I think that's because my opam2.1 setup used a different `OPAMROOT` and this setting was ignored.

PR #195, opened 2021-08-26
Commit_pointed_by: partial support when ref is a commit

TheLortex:
Fixes https://github.com/ocamllabs/opam-monorepo/issues/127

In the ref resolution we can check if a commit exists when it's pointed by an existing ref. 
Otherwise we just check if it looks like a commit (hex numbers), assume it exists in the remote and emit a warning.

### DIFF
--- a/lib/git.ml
+++ b/lib/git.ml
@@ -34,9 +34,22 @@ module Ls_remote = struct
     in
     List.fold_left ~f ~init:{ maybe_packed = None; not_packed = None } lines
 
+  let is_commit target = List.exists ~f:(fun (commit, _) -> commit = target)
+
+  let looks_like_commit ref =
+    String.length ref > 6
+    && Astring.(String.for_all Char.Ascii.is_hex_digit ref)
+
   let commit_pointed_by ~ref output_lines =
+    let log_approx () =
+      Logs.debug (fun l ->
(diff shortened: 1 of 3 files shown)


lib/git.ml
```diff
@@ -34,9 +34,22 @@ module Ls_remote = struct
...
 
+  let is_commit target = List.exists ~f:(fun (commit, _) -> commit = target)
+
+  let looks_like_commit ref =
+    String.length ref > 6
+    && Astring.(String.for_all Char.Ascii.is_hex_digit ref)
+
   let commit_pointed_by ~ref output_lines =
+    let warn_approx () =
+      Logs.warn (fun l ->
```
NathanReb, line 45: How about making this a debug log instead?

I think there are valid use cases for pinning to a commit that's not the head of any branch or isn't tagged and users that rely on it will have to ignore these warnings. Making it a debug log would make it invisible by default.

NathanReb (commented):
This looks good! Thanks for taking care of this!

I just have a minor comment regarding the warning but otherwise this is good to go!

NathanReb:
Can you also add a changelog entry about this?

TheLortex:
Thank you for the review ! I've added a changelog entry and updated the code to use `Logs.debug` instead of `Logs.warn`.

NathanReb (approved):

PR #196, opened 2021-08-31
Add lwt dep.

rizo:
`lwt` is required on a clean installation.


Leonidas-from-XIV:
Looking at the code it doesn't look like it uses Lwt at all, so I think it would be more sensible to remove that spurious dependency.

PR #198, opened 2021-09-20
Set the `root_dir` in the OPAM state

### LINKED ISSUE #197: Fails for when OPAMROOT is set to something other than the default directory
When `OPAMROOT` is set to something other than the default directory, `opam-monorepo` fails:

```
[ERROR] Opam has not been initialised, please run `opam init'
               OpamStd.OpamSys.Exit(50)
               Raised at file "src/core/opamStd.ml", line 1042, characters 15-29
               Called from file "src/state/opamGlobalState.ml", line 60, characters 4-113
               Called from file "src/state/opamGlobalState.ml", line 185, characters 11-20
               Called from file "cmdliner_term.ml", line 25, characters 19-24
               Called from file "cmdliner.ml", line 33, characters 27-34
               Called from file "cmdliner.ml", line 117, characters 32-39
make: *** [Makefile:61: deps-pull] Error 125
```

It doesn't seem like monorepo is initialising the opam state in the same way opam does, so the `OPAMROOT` variable is never read.

Leonidas-from-XIV:
If not set then OPAM does not set the right path for the OPAM root and will use the default `~/.opam` no matter what. The OPAM API is surprising in this regard that even passing the root folder to `load_defaults` does not actually end up setting the folder.

Closes #197

### DIFF
--- a/cli/dune
+++ b/cli/dune
@@ -1,4 +1,12 @@
 (library
  (name duniverse_cli)
- (libraries duniverse_lib fmt.cli fmt.tty logs.cli logs.fmt cmdliner lwt.unix
-   dune-build-info ocaml-version))
+ (libraries
+  duniverse_lib
+  fmt.cli
+  fmt.tty
+  logs.cli
+  logs.fmt
+  cmdliner
+  lwt.unix
+  dune-build-info
+  ocaml-version))
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,3 +1,17 @@
+## unreleased
+
+### Added
+
+### Changed
+
+### Deprecated
+
+### Fixed
+
+### Removed
+
+### Security
+
 ## 0.2.4
 
 ### Added
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,9 @@
 
 ### Fixed
 
+- Fix setting `OPAMROOT` to accept non-default paths (#197, #198,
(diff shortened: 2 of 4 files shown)


emillon:
Thanks!

cli/dune
```diff
@@ -1,4 +1,12 @@
 (library
  (name duniverse_cli)
- (libraries duniverse_lib fmt.cli fmt.tty logs.cli logs.fmt cmdliner lwt.unix
-   dune-build-info ocaml-version))
+ (libraries
+  duniverse_lib
```
NathanReb, line 4: This formatting change should have triggered a CI failure. Trying to format the repo locally reformats them as before.

Leonidas-from-XIV: That's strange. Given dune reformats its own files, this shouldn't have anything to do with ocamlformat. So it must be the dune version. I just tried on `main` (with this merged) and dune 2.9.1 keeps them. Maybe some change in the dune formatting defaults?

emillon: Something changed but these changes are tied to `lang dune`. So if your editor integration uses `dune format-dune-file` instead of `build @fmt` it might pick the wrong version. And CI for opam-monorepo projects does not have a lint-fmt build at the moment so it's not checked in CI.

emillon: I talked too fast, there's a lint-fmt build now apparently

PR #199, opened 2021-09-27
Add build badge to README

NathanReb:


Leonidas-from-XIV (approved):
TIL about shields.io that it supports arbitrary systems to make badges. Cool.

PR #201, opened 2021-09-27
Remove Lwt dependency

Leonidas-from-XIV:
We don't use Lwt seemingly anywhere.

Closes #196.


NathanReb (approved):
Thanks! Looks good!

Let's wait until the CI comes back and then we can merge!

PR #203, opened 2021-10-11
Filter local from lock

Leonidas-from-XIV:


cli/lock.ml
```diff
@@ -209,20 +171,107 @@ let get_pin_depends ~global_state local_opam_files =
...
+
+let subselect_explicit ~local_packages ~explicit_list =
+  let by_opam_name entry (explicit_name, _) =
+    let name = entry |> Package_argument.name |> OpamPackage.Name.to_string in
+    String.equal name explicit_name
+  in
+
+  List.fold_left
+    ~f:(fun acc explicit ->
+      match (List.find_opt ~f:(by_opam_name explicit) local_packages, acc) with
```
NathanReb, line 191: This could probably be improved by using a local packages map, which we're build after that!

cli/lock.ml
```diff
@@ -209,20 +171,107 @@ let get_pin_depends ~global_state local_opam_files =
...
+  |> Result.map_error ~f:(fun missing_packages ->
+         let msg =
+           Fmt.str "Packages %a specified but not found in repository"
+             Fmt.(list ~sep:comma Package_argument.pp_styled)
+             missing_packages
+         in
+         `Msg msg)
+
+let only_return_root xs =
+  let root = Sys.getcwd () |> Fpath.v in
```
NathanReb, line 206: We shouldn't use `getcwd` here but rather use the `repo` argument. It defaults to `cwd` but can be overwritten using a CLI option.

cli/lock.ml
```diff
@@ -209,20 +171,107 @@ let get_pin_depends ~global_state local_opam_files =
...
+      match Fpath.relativize ~root path with
+      | None -> false
+      | Some path ->
+          let filename = path |> Fpath.filename |> Fpath.v in
+          Fpath.equal filename path)
+    xs
+
+let construct_version (name, repo) = (name, (None, repo))
+
+let group ~equal xs =
```
NathanReb, line 218: If we end up using this function it would better fit in our stdext.

cli/lock.ml
```diff
@@ -209,20 +171,107 @@ let get_pin_depends ~global_state local_opam_files =
...
+        names);
+  unique
+
+let deduplicate_packages l =
+  let cmp (a, _) (b, _) = String.compare a b in
+  let equal (a, _) (b, _) = String.equal a b in
+  List.sort ~cmp l |> group ~equal |> filter_unique |> List.concat
+
+let target_path_map packages =
+  packages |> deduplicate_packages |> String.Map.of_list_exn
```
NathanReb, line 248: It seems a bit weird to operate the deduplication on the list and to build a map after that. This can be done while building the map itself using something like `Map.update`.

Leonidas-from-XIV: I thought about this but it is dependant on the semantics. If we want to remove all duplicated packages instead of just ignoring everything past the first one we encounter then this has to be done separately since otherwise we'll get into an odd edge case:

1. Package `a` found. We add it to the map
2. Another package `a` found. We attempt to add it to the map, realize it is already there so we remove both.
3. Another package `a` found. The map had its previous `a` removed, so we think this is the first time we see it, so we remove it.

This could be done with an additional set to track what we have already excluded but that feels to me quite roundabout, whereas this way we get all the duplicated packages in a data structure and can print them in one go so the user can get good feedback on what is wrong.

The code here is not great though, admittedly, with `deduplicate_packages` just doing some side-effect willy-nilly.

NathanReb: I think a good behaviour is to keep at most one of each packages, meaning if there is a duplicate, we keep the first one we encountered, report about the others. Accumulating both a map and a list of ignored duplicates doesn't sound too bad to me.
It would have the advantage that we can report duplicates all at once in a more helpful manner, something like:
```
Package x is defined multiple times in the repo:
- ./path/to/first/x.opam
- ./path/to/second/x.opam
We kept ./path/to/first/x.opam and discarded the others
```

How does that sound?

We can still improve this in the future when we have an idea if this actually shows up and why!

I think in the end the proper fix will be to use some dune magic to know which packages are actually considered by dune at which point we can safely reject the repo as `dune build` won't work anyway.

I'm digressing a bit but my point is that there is no simple yet perfect behaviour here so let's start with the simplest one and we'll improve it overtime when we have a better idea about what the problem actually is.

cli/lock.ml
Leonidas-from-XIV: You're right. I removed this and moved/rewrote code to use package-name keyed `String.Map.t`s now.

NathanReb (commented):
It's getting better but I think there are still things that aren't interpreted properly and some extra unnecessary computations. Looking at the flow  in the main function in `Lock`, it seems to be overly complex, allocating a lot of redundant data structures and I admit I'm not entirely sure it's doing the right thing as I'm having a bit of troubles following it.

I'll try to sum up what data structures we need, in the order in which we need them, considering the current API we have for `Opam_solve`. It should clarify what the simplest flow should be:
1. We need a Map representing all local packages. That's the `local_opam_files` argument of the solver. This depends on the `repo` alone, nothing else. It contains mappings from package names to a single pair `version * opamfile`.
2. We need a set of package names that will be our target packages, used for the solver request and to collect some other bits of information such as what is called the `root_depexts` atm. It's the `target_packages` argument to the solver. It is built from the CLI argument `target_packages`, the CLI flag `recurse` and the map built in `1`, which we need to ensure they actually exist locally.

I might be wrong but I think any other data structure allocated here is superfluous. At the moment we create a `targetted_paths` map which is only used to extract a set of package names and to build `target_opam_files` which is redundant with `local_opam_files`.
For example the `root_depexts` can be computed from just `1` and `2` by looking up the names from `2` in the map `1` and extracting the depexts from the corresponding opam files.

I think a key point is that `1` is independent from `2`, in the sense that we don't need to know what `2` is to assemble `1`.
Getting rid of other intermediate data structures from the main function would help understanding the flow but I think would also help get rid of redundancy here. I think as a rule of thumb, we shouldn't build a variable in the main function if it's not used to do anything meaningful (running the solver, assembling the lockfile, etc...).

I didn't mention the extra step that is setting the explicit versions but that's just editting the data structure `1` using the `specified_packages`. Basically when building `1` initially we can map everything to the version we already map them to (that is the one in the opam file or `zdev` if there's none) and then overwrite that value with the one provided on the command line, if any, in the specified packages.

I'm happy to discuss this over a call, I understand I might not have provided enough information here and writing a bit more detailed spec could have helped.

Let me know what you think and if you need any help with this!

NathanReb:
I tested this locally, using [NathanReb/test-multi-opam-monorepo](https://github.com/NathanReb/test-multi-opam-monorepo).

Trying to run `opam monorepo lock` on that repo will fail either because it doesn't know about `b` and `c` or, if you pin them to the remote repo, pull will fetch a copy of the repo in the duniverse and `dune build` will fail because of packages being defined more than once.

It works with you branch. The local packages, even the non target ones are not present in the lockfile but I think that's what we wanted and after thinking about this further it does make sense this way, especially if we wanted to opam install the lockfile to be able to build our target packages (we don't necessarily want opam to install our local packages).

I tested the following scenarios successfully:

1. `opam monorepo lock --recurse`, targets are `a` `b` and `c`. That's the only scenario that works on main right now.
2. `opam monorepo lock`, targets are `a` and `b`, `c` is properly handled as local package, target packages are still the expectd ones, defined at the root
3. `opam monorepo lock a`, target is `a`, `b` and `c` are still properly handled as local packages

*At this point I added a `{= version}` constraint to the `a -> b` dependency:*

4. `opam monorepo lock a.1.0.0`, it fails as expected because `b` is assigned `zdev` and therefore the solver can't find a solution
5. `opam monorepo lock a.1.0.0 b.1.0.0` fixes the above problem and works properly
6. `opam monorepo lock a.1.0.0` after adding a version field in `b.opam` with `1.0.0`, it also fixes the issue in `4` as expected

If you have any good test scenarios I'm happy to try them out!

NathanReb (approved):
Looks good to me! It would be nice to test this on Tezos, cc @emillon but I think the tests I've been running are a good starting point and that this can be merged as is, unless you come up with good test scenari that you'd want to try first!

NathanReb:
Can we merge this?

Leonidas-from-XIV:
I think this is ok to merge. If we find additional issues we can handle this as bugfixes and will have reproducible cases where it fails.

PR #200, opened 2021-09-27
Upgrade ocamlformat

NathanReb:

PR #204, opened 2021-10-11
Upgrade lockfile and ocamlformat

NathanReb:
This upgrades our lockfile to the latest versions of our deps, in particular it upgrades the opam libs to actual 2.1 instead of the `rc2` versions.


NathanReb:
I merged #200 into this hoping it will help us work around our CI issues!

PR #209, opened 2021-10-18
Move lockfiles extracting function to Repo

NathanReb:
With the fix to the lockfile path inference, we now allow users to have several lockfiles in their repositories.

This function will likely prove more and more useful so I think it makes sense to have it here.

As a follow up I'll look into eventually making the depext subcommand install depexts for all local lockfiles.

### DIFF
--- a/cli/common.ml
+++ b/cli/common.ml
@@ -106,23 +106,12 @@ let filter_duniverse ~to_consider (duniverse : Duniverse.t) =
             Fmt.(list ~sep string)
             unmatched)
 
-let root_lockfiles repo =
-  let open Result.O in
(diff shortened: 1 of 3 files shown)


emillon:
> As a follow up I'll look into eventually making the depext subcommand install depexts for all local lockfiles.

I'm not sure that's correct in all cases. For example if several lockfiles correspond to different configurations, installing all depexts is not necessarily the right thing to do. `ocaml-ci` won't mind though, since when it's installing depexts it has only one lockfile in the source tree. But as long as we can pass the lockfile path explicitly we're fine I think.

PR #217, opened 2021-10-21
Add opam-monorepo list

samoht:
The output is similar to `opam list` but reading the current lockfile

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,7 @@
 
 ### Added
 
-- Add a list subcommand to list the packages available in the monorepo
+- Add a list subcommand to list the duniverse packages in the lockfile
   (#217, @samoht)
 
 ### Changed


NathanReb (commented):
Sorry for the late review!

This looks good! I'm not 100% sold on the "guess pin" idea but we can always come back to this later.

Besides that I have a few minor comments but then we should be good to go!

cli/list.ml
```diff
@@ -0,0 +1,136 @@
...
+        match t.url with Git _ -> true | _ -> guess_pin ~version ~loc
+      in
+      let descr = None in
+      { name; version; descr; pinned; loc })
+    t.provided_packages
+
+let pkgs_of_duniverse t =
+  let pkgs = List.map ~f:pkgs_of_repo t in
+  let pkgs = List.flatten pkgs in
+  (* FIXME: why do we have duplicates here ? *)
```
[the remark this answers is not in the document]
samoht: all of the entries were duplicated ; I don't know why :-) maybe because they were both in the `depends` and `pin-depends` fields?

cli/common.ml
```diff
@@ -104,7 +104,7 @@ let filter_duniverse ~to_consider (duniverse : Duniverse.t) =
         String.Map.of_list_map_exn duniverse ~f:(fun src -> (src.dir, src))
       in
       let unmatched, found =
-        List.partition_map to_consider ~f:(fun asked ->
+        Import.List.partition_map to_consider ~f:(fun asked ->
```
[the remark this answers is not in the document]
NathanReb: Ok this is due to ocamldep mistaking `List` from `Import` with the `List` from `Duniverse_cli`. It's a bit annoying that there's no other way around that but I renamed the module in `Duniverse_cli` as we use `List` quite a lot around these parts!

cli/list.ml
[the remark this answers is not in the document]
NathanReb: That sounds like a bug in the `Lockfile` -> `Duniverse.t` conversion, I'll investigate!

NathanReb:
I rebased and pushed some small improvements here and there. I'll investigate the duplicate thing and will get back to this!

NathanReb:
I finished polishing this as @samoht did not have time to work on this recently. A quick, final review would be nice!

cli/list_cmd.ml
```diff
@@ -0,0 +1,138 @@
...
+  name : string;
+  version : string;
+  loc : string;
+  pinned : bool;
+  descr : string option;
+}
+
+(* Assumes max_len was correctly computed, otherwise the input string might be
+   truncated. *)
+let pad s max_len =
```
emillon, line 14: I think that this can be expressed as:

```ocaml
Printf.sprintf "%-*s" max_len s
```

(this also assumes that `max_len` is correct - if `s` is larger it's going to overflow)

cli/list_cmd.ml
```diff
@@ -0,0 +1,138 @@
...
+  let len = String.length s in
+  String.init max_len ~f:(fun i -> if i < len then s.[i] else ' ')
+
+(* FIXME: replace this once we remember the remotes used during lock. *)
```
emillon, line 18: If this can be turned into a pointer to an issue that's more probability that it's going to be addressed :)

NathanReb: I'll simply remove this one as I'm not sure we'll ever remember remotes.

cli/list_cmd.ml
NathanReb: That's neat! A bit cryptic I think but I guess the wrapping function name should be enough.

PR #227, opened 2021-11-05
Properly handle missing dune-project

### LINKED ISSUE #225: Running opam-monorepo in a project without `dune-project` file at the root crashes
It does it the following way:
```
$ opam monorepo lock
==> Using 2 locally scanned packages as the roots.
               (Sys_error
                 "/home/nathan/code/test-multi-opam-monorepo/dune-project: No such file or directory")
               Raised by primitive operation at Stdlib.open_in_gen in file "stdlib.ml", line 399, characters 28-54
               Called from Stdlib.open_in in file "stdlib.ml" (inlined), line 404, characters 2-45
               Called from Sexplib__Pre_sexp.gen_load_rev_sexps in file "duniverse/sexplib/src/pre_sexp.ml", line 688, characters 11-23
               Called from Sexplib__Pre_sexp.load_rev_sexps in file "duniverse/sexplib/src/pre_sexp.ml" (inlined), line 699, characters 31-75
               Called from Sexplib__Pre_sexp.load_sexps in file "duniverse/sexplib/src/pre_sexp.ml" (inlined), line 700, characters 36-62
               Called from Duniverse_lib__Dune_file.Raw.as_sexps in file "lib/dune_file.ml", line 73, characters 11-59
               Called from Duniverse_lib__Repo.project_name in file "lib/repo.ml", line 55, characters 2-37
               Called from Duniverse_lib__Repo.lockfile.(fun) in file "lib/repo.ml", line 74, characters 9-23
               Called from Duniverse_cli__Lock.run.(fun) in file "cli/lock.ml", line 233, characters 2-55
               Called from Cmdliner_term.app.(fun) in file "duniverse/cmdliner/src/cmdliner_term.ml", line 25, characters 19-24
               Called from Cmdliner.Term.term_result.(fun) in file "duniverse/cmdliner/src/cmdliner.ml", line 33, characters 27-34
               Called from Cmdliner.Term.run in file "duniverse/cmdliner/src/cmdliner.ml", line 117, characters 32-39
```

Exitting gracefully while still complaining about the project not having a root dune-project is the least (and best) we can do here!

NathanReb:
Fixes #225 !

Depends on #203.

This PR improves the error message when a dune-project file is missing at the root. As the test shows, it now prints a nice error with suggested actions instead of an uncaught exception with a backtrace!


Leonidas-from-XIV:
Just merged #203, can you please rebase? That will make reviewing a bit easier.

lib/project.mli
```diff
@@ -24,16 +24,16 @@ val name : t -> (string, [> `Msg of string ]) result
...
+(** Returns the path to the opam-monorepo lockfile to generate for the given
+    project and lockfile target packages.
+    If there is a single target package, then it is the ["<package_name>.opam.locked"]
     file at the root of the project.
     If it contains multiple packages, then it's the ["<project_name>.opam.locked"] file
-    at the root of the project.
-    One can provide [local_packages] if they were already computed or if only a subset
-    of the local packages must be taken into account. *)
+    at the root of the project, where <project_name> is the name as defined in the
+    dune-project file. *)
```
Leonidas-from-XIV, line 36: Given the above documentation string is taking quite some lengths to describe the name of the lockfile with markup, how about this:

```suggestion
    at the root of the project, where [<project_name>] is the name as defined in the
    [dune-project] file. *)
```

test/bin/missing-dune-project/run.t
```diff
@@ -0,0 +1,34 @@
...
+                 Called from Sexplib__Pre_sexp.load_rev_sexps in file "duniverse/sexplib/src/pre_sexp.ml" (inlined), line 699, characters 31-75
+                 Called from Sexplib__Pre_sexp.load_sexps in file "duniverse/sexplib/src/pre_sexp.ml", line 700, characters 36-62
+                 Called from Duniverse_lib__Dune_file.Raw.as_sexps in file "lib/dune_file.ml", line 73, characters 11-59
+                 Called from Duniverse_lib__Project.name in file "lib/project.ml", line 40, characters 2-37
+                 Called from Duniverse_lib__Project.lockfile in file "lib/project.ml", line 50, characters 9-15
+                 Called from Duniverse_cli__Lock.run.(fun) in file "cli/lock.ml", line 310, characters 2-56
+                 Called from Cmdliner_term.app.(fun) in file "duniverse/cmdliner/src/cmdliner_term.ml", line 25, characters 19-24
+                 Called from Cmdliner.Term.term_result.(fun) in file "duniverse/cmdliner/src/cmdliner.ml", line 33, characters 27-34
+                 Called from Cmdliner.Term.run in file "duniverse/cmdliner/src/cmdliner.ml", line 117, characters 32-39
+  [125]
```
Leonidas-from-XIV: I greatly appreciate the adding of an `cram` test, because I think being able to test the existing and desired behavior of `opam-monorepo` is great but I feel that printing this exception will be extremely fragile and lead to breakages everywhere.

I was thinking on how to improve this without losing the `125` status code when using pipes. I don't know if `cram` supports something like `[125|0]` (e.g. displaying the status codes of pipe components) but if not it could be possible to `2> stderr` and then `grep ... < stderr` to make the output a bit more reproducible.

CHANGES.md
```diff
@@ -16,6 +16,11 @@
 
 - Better errors for `opam-monorepo depext`, especially for non-interactive
   shells (#216, @samoht)
+- Properly detect all opam packages defined in the current repository, preventing it
+  from later pulling duplicates into the duniverse if they were part of the target packages
+  dependencies. (#203, @Leonidas-from-XIV)
+- Properly report missing dune-project file when trying to determine the
+  to-be-genrated lockfile name (#227, @NathanReb)
```
Leonidas-from-XIV, line 23: Tiny typo:

```suggestion
- Properly report missing `dune-project` file when trying to determine the
  to-be-generated lockfile name (#227, @NathanReb)
```

Leonidas-from-XIV (changes requested):
This definitely brings `opam-monorepo` closer to where we want to be. I'm actually more excited about the first `cram` test than then fix itself. But we need to make sure the test output is less fragile.

cli/lock.ml
```diff
@@ -137,6 +137,12 @@ let lockfile_path ~explicit_lockfile ~target_packages repo =
       Project.lockfile
         ~target_packages:(OpamPackage.Name.Set.elements target_packages)
         repo
+      |> Result.map_error ~f:(function `Msg msg ->
+             Rresult.R.msgf
+               "Could not infer the target lockfile name: %s\n\
+                Try setting it explicitly using --lockfile or add a project \
```
Leonidas-from-XIV, line 143: I'm wondering whether it wouldn't be nice to use some `Fmt.styled` option for `--lockfile` to mark it up as a shell command. It is a small thing. What do you think?

test/bin/missing-dune-project/run.t
NathanReb: You'll note that this is not the cram test we're merging. This was added before the fix so that the history demonstrate the existing bug and corresponding fix.

cli/lock.ml
NathanReb: Yup definitely make sense to improve this, I'll give it a spin. I'm not sure how well that works with the intermediate string allocated.

Might actually be a better idea to log this ourselves and `exit 1` here rather than embeding color and style encoding into a string that's carried around.

lib/project.mli
Leonidas-from-XIV: Ah. I always thought these served the same way as ``` ` ``` do in Markdown.

cli/lock.ml
Leonidas-from-XIV: Come to think about it, you're probably right. This is more a `panic` (to use Rust lingo) level output from which recovery is impossible, carrying around a colorful string does not give much value.

OTOH, terminating the program from a random place feels a bit dirty. I wonder whether it wouldn't make sense to make this a polymorphic variant like ``` `Unrecoverable of string```. Thought that's probably overcomplicating it.

Leonidas-from-XIV (approved):
With the latest commit the tests look fine to me. One can consider doing something with the error reporting to exit directly instead but I wouldn't delay this PR to find the most theoretically beautiful solution.

cli/lock.ml
NathanReb: The formatting cannot be embedded in the string, I just tried it.

I agree terminating the program isn't great. One thing worth considering is that we could log the error there and let the main function termintate it. We could indeed use a special variant for this so that we're sure it's not intercepted. One that would have no attached message could be nice as well since we can assume that the error was reported already. If the program is to be terminated there is no reason to carry around the error message anymore.

How does that sound?

lib/project.mli
emillon: I think that "raw" syntax is `{v ... v}`.

test/bin/missing-dune-project/run.t
```diff
@@ -0,0 +1,21 @@
+Let us setup a simple project with two local packages, defined at the root
```
emillon, line 1: Since this test is self-contained (you create necessary files in the test itself) you can just name it `test/bin/missing-dune-project.t`.

Don't forget to move the dune file up!

test/bin/missing-dune-project/dune
```diff
@@ -0,0 +1,2 @@
+(cram
+ (deps %{bin:opam-monorepo}))
```
emillon, line 2: If you move this file up and add `(applies_to :whole_subtree)` it'll apply to all future cram tests.

test/bin/missing-dune-project/run.t
NathanReb: I'm not sure I'd want to mix files and directories here. Having a directory as the adavantage to be more future proof here as for more complex test cases, it might be good to have a directory inside which you CD and run the command yourself.

test/bin/missing-dune-project/dune
NathanReb: Good point! I'll move it!

cli/lock.ml
NathanReb: After giving this a quick spin, it's not as easy to deal with so let's leave it out of this PR.

It would be nice to have a proper way to deal with this but it requires more in depth changes and those would be better done on their own!

test/bin/missing-dune-project/run.t
NathanReb: I read the documentation for cram tests and I'm a bit surprised they went for this. It seems that if you don't use the `dir.t/` feature you'll have to declare the dependencies of your test by hand so the best solution, in terms of consistency, would be to put all tests in a `test-name.t/` folder, whether they have extra files or not.

I also wonder if it wouldn't make more sense to always put the files we need in the directory rather than generating them on the fly as part of the test.

NathanReb: @Leonidas-from-XIV any thoughts on this?

NathanReb:
I moved the test to a cram folder an wrote the file to the source tree instead of generating them. I think that's a better practice for writing tests.

I still `cat` the files in `run.t` so that it's self contained and no file browsing is required to understand the test.

`cd`ing inside the test folder and running `dune exec -- opam-monorepo ...` now makes for a better dev experience if one wants to reproduce or debug.

emillon (approved):

test/bin/missing-dune-project/run.t
Leonidas-from-XIV: @NathanReb I think it would indeed be a bit nicer to have the files in there, because the tests often spend quite some (extremely repetitive) lines to create the content.

What I am less happy about is that files like `.gitignore` would be there too and picked up by the "outer" project git, thus effectively often ignoring `run.t` which we actually don't want to have ignored. As a compromise, the `.gitignore` file could be written/`mv`'d inside the test. I am not amazed by that idea, but maybe it is a viable 80% solution? What do you think?

NathanReb: Yeah that sounds good!

PR #231, opened 2021-11-10
Update lockfile

NathanReb:
This bumps our lockfile and pins opam-0install to its master branch as the pacth we rely one was merged but not yet released.

### DIFF
--- a/opam-monorepo.opam
+++ b/opam-monorepo.opam
@@ -40,7 +40,7 @@ build: [ "dune" "build" "-p" name "-j" jobs "@install" "@runtest" {with-test} ]
 flags: [ plugin ]
 pin-depends: [
   [
-    "opam-0install.0.4.2"
-    "git+https://github.com/NathanReb/opam-0install-solver#0e3d946084d6da0aa33f0a0c61e0cc780025054e"
+    "opam-0install.dev"
+    "git+https://github.com/ocaml-opam/opam-0install-solver"
   ]
 ]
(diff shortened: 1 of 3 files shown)


emillon (approved):

opam-monorepo.opam
```diff
@@ -40,7 +40,7 @@ build: [ "dune" "build" "-p" name "-j" jobs "@install" "@runtest" {with-test} ]
 flags: [ plugin ]
 pin-depends: [
   [
-    "opam-0install.0.4.2"
-    "git+https://github.com/NathanReb/opam-0install-solver#0e3d946084d6da0aa33f0a0c61e0cc780025054e"
+    "opam-0install.dev"
```
Leonidas-from-XIV, line 43: I'm wondering whether it wouldn't be better to update to 4eede964fabba756e4c3c825f15da3bb499a9356 (the merged diagnostics) or c5b44d7b6f4aef3585e59b190efb3a1de343682c (current HEAD as of today) as long as there isn't any new release that we would use as minimal version.

NathanReb: I think what we want to say here is that we rely on the dev version, not necessarily on that specific commit. E.g. if a new feature is merged to the main branch, relocking will allow us to benefit from it, same as a new release would without the pin.

The lockfile contains a fixed commit to make sure we're all on the same page when developing opam-monorepo but I'm not sure it makes sense to pin to a commit. That would be equivalent to saying `opam-0install {= "0.4.2"}` in our opam file in the sense that we explicitly don't want a higher version to be selected if it's available. I don't think it's the case here.

What do you think?

Leonidas-from-XIV: You're probably right. The think I like about adding a commit is that it announces "it has to be at least this version". In the lockfile this is taken care of by locking to a specific commit but the `.opam` file is a bit laissez-faire. But I guess this is more a theoretical concern and once opam-0install-solver has a release this will be resolved anyway.

PR #232, opened 2021-11-10
Clarify which opam-monorepo to use in opam-monorepo

NathanReb:
It might be tempting to use opam-monorepo on itself but the latest released version should be used instead, to make sure
the CI can keep functioning at all time and that new, potentially untested features don't break our dev workflow.


emillon (approved):

Leonidas-from-XIV (approved):
:+1:

Maybe we should also make via an action that when the opam file is modified, the lock file is updated as well?

PR #233, opened 2021-11-10
Fix a recent bug where pin-depends would be excluded form the lockfile

NathanReb:
The solver was running properly, with the pins enabled but the packages appearing in a pin-depends field of one of the local packages were filtered out of the final solution.


NathanReb:
I'm adding a no-changelog label as this wasn't released!

Leonidas-from-XIV (approved):
Makes sense for this to be a "no changelog" change because it wasn't observed by any users.

PR #234, opened 2021-11-12
Allow marking certain packages as "installed via OPAM"

Leonidas-from-XIV:
This is a draft PR on the hybrid mode feature, which makes it possible to actually test things and go somewhere from there.

1. Packages to be installed via `opam` can't be marked as not-installable in the solver because in such case there is no solution. I explored this avenue and it doesn't lead anywhere. At least one part of the solution space eliminated :heavy_check_mark: 
1. There's currently warnings about `opam-provided`, our custom variable. Not great.
1. Variables are false by default so packages marked this way will not be picked up by opam.
1. The detection of the variable is extremely inflexible now. This is just (very necessary) busywork to walk that formula and determine whether the variable is set but needs to be done before release. For testing the concept on examples this works ok. 
1. It needs a command to install the `opam-provided` packages. That's the easiest part so I'm leaving it for last, once we have figured out the other problems just collecting packages to install should be simple.

(No changelog entry because it shouldn't be merged and also for a decent changelog entry I'd want to have the functionality in place so I know what to describe)

But what works is:
1. Mark a package as `opam-provided` in the `opam`/`dune-project` file, e.g. `bos` like in this PR
1. Run `opam-monorepo lock`
1. See that `bos` and its dependencies like `fpath`, `rresult` and friends aren't part of the lockfile anymore, so they wouldn't be pulled anymore.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,17 @@ opam-repository) thanks to the `pin-depends`.
 You can use that property to your advantage by allowing one to choose between a "monorepo" or
 regular opam workflow depending on the situation.
 
+You can also exclude packages from being included into the set of packages to
+be vendored by opam-monorepo. To do so you can specify an additional field in
+your OPAM file:
+
+```
+x-opam-monorepo-opam-provided: ["ocamlformat" "patdiff"]
+```
+
+This will exclude the packages from the list of packages opam-monorepo will
+pull, so they can be installed via `opam` manually.
+
 ### opam monorepo pull
 
 The `pull` command fetches the sources using the URLs in the lockfile. It benefits from the opam
--- a/README.md
+++ b/README.md
@@ -79,20 +79,20 @@ opam-repository) thanks to the `pin-depends`.
 You can use that property to your advantage by allowing one to choose between a "monorepo" or
 regular opam workflow depending on the situation.
 
-You can also exclude packages from being included into the set of packages to
-be vendored by opam-monorepo. To do so you can specify an additional field in
-your OPAM file:
+You can also exclude packages from the set of packages to
+be vendored by `opam-monorepo`. To do so you can specify an additional field in
+your Opam file:
 
 ```
 x-opam-monorepo-opam-provided: ["ocamlformat" "patdiff"]
 ```
 
-This will exclude the packages from the list of packages opam-monorepo will
+This will exclude the packages from the list of packages `opam-monorepo` will
 pull, so they can be installed via `opam` manually.
 
 ### opam monorepo pull
 
-The `pull` command fetches the sources using the URLs in the lockfile. It benefits from the opam
+The `pull` command fetches the sources using the URLs in the lockfile. It benefits from the Opam
 cache but its outcome does not depend on your opam configuration.
 
 ## Monorepo projects
--- a/doc/opam-provided.mld
+++ b/doc/opam-provided.mld
@@ -1,15 +1,15 @@
 {1 [opam]-provided dependencies}
 
-This section documents a feature that is useful if not all dependencies can be
-installed via [opam-monorepo]. This workflow is not meant for regular use but
-only as a way to enable to use [opam-monorepo] in cases where it wouldn't be
+This section documents a feature that’s useful when some dependencies cannot be
+installed via [opam-monorepo]. This workflow isn’t meant for regular use but
+only as a way to use [opam-monorepo] in cases where it wouldn't be
 otherwise possible.
 
-We suggest minimizing its usage and prefer to use (and contribute) dune-ports
+We suggest minimizing its usage and prefer to use (and contribute) [dune-ports]
 from the [opam-overlays] repository. It's usage is safest with
-leaf-dependencies and those that do not have no dependency intersections with
-vendored packages. Due to possibly unexpected interactions this feature is only
-meant for advanced users and the solutions that it produces might be in flux.
+leaf-dependencies and those that don’t have dependency intersections with
+vendored packages. Due to possibly of unexpected interactions, this feature is only
+meant for advanced users, and the solutions it produces might be in flux.
 
 {2 Usecases}
 
@@ -18,43 +18,43 @@ be it due to your dependencies not building with [dune] or some tooling that
 should be just installed via [opam]. Such examples include:
 
 {ul
-{- Packages not building with dune}
+{- Packages not building with Dune}
 {- Packages that should not be vendored for legal reasons}
 {- Packages that just provide helper binaries like [utop], [ocamlformat]}
 }
 
-These users can opt-in into having some of their dependencies provided by
-[opam] instead of [opam-monorepo] pulling them into the duniverse.
+These users can opt-in to having some of their dependencies provided by
+[opam] instead of [opam-monorepo] pulling them into the [duniverse].
 
 {2 Initial setup}
 
-With the default configuration [opam-monorepo] runs in full-duniverse mode,
-that is requiring all the dependencies of your [opam] files to be able to be
-built as part of the duniverse.
+With the default configuration, [opam-monorepo] runs in full-[duniverse] mode,
+i.e., requiring all the dependencies of your [opam] files to be able to be
+built as part of the [duniverse].
 
 As such it is only recommended that packages are installed via [opam] for which
-there is no way to install them via [dune], e.g. by using the
+there is no way to install them via [dune], e.g., by using the
 {{:https://github.com/dune-universe/opam-overlays}opam-overlays} repository
-with dune ports.
+with Dune ports.
 
-To let [opam-monorepo] know a package is to be installed via [opam] it needs to
-be marked as such in the opam file. For this the
-[x-opam-monorepo-opam-provided] field is used, which takes a single or a list
-of packages to be installed via [opam] instead of pulled into the duniverse.
+To let [opam-monorepo] know a package is to be installed via [opam], it needs to
+be marked as such in the Opam file. For this, we use the
+[x-opam-monorepo-opam-provided] field, which takes a single or a list
+of packages to be installed via [opam] instead of pulled into the [duniverse].
 
-To configure [opam-monorepo] to avoid pulling in package [foo] specify it in
-the list of packages that are to be provided by [opam]:
+To configure [opam-monorepo] to avoid pulling in package [foo], specify it in
+the list of packages provided by Opam:
 
 {[
 x-opam-monorepo-opam-provided: ["foo"]
 ]}
 
-This can either be specified directly in the [opam] file or, if you use [dune]
-to generate [opam] files, in the [.template] file. In such case make sure to
-re-generate the [opam] file.
+This can either be specified directly in the Opam file or, if you use Dune
+to generate Opam files, in the [.template] file. In such case, make sure to
+regenerate the Opam file.
 
-As a shortcut syntax if there is only one package it is possible to leave the
-list out and just specify the package itself:
+As a shortcut syntax, if there’s only one package, it’s possible to leave out the
+list and just specify the package itself:
 
 {[
 x-opam-monorepo-opam-provided: "foo"
@@ -62,15 +62,14 @@ x-opam-monorepo-opam-provided: "foo"
 
 {2 Usage}
 
-After you configured your [opam] file, you have to {e lock} the environment.
+After you configured your Opam file, you have to {e lock} the environment.
 
 {[
 $ opam monorepo lock
 ]}
 
-After this succeeds, you will have a [.opam.locked] file, which contains all
-the dependencies (including transitive dependencies) of your project. The ones
-that [opam-monorepo] will pull into the duniverse are marked as [vendor].
+After this succeeds, you will have an [.opam.locked] file, which contains all your project’s dependencies (including transitive dependencies). The ones
+that [opam-monorepo] will pull into the [duniverse] are marked as [vendor].
 
 {[
 $ opam install --ignore-pin-depends --deps-only ./ --locked
@@ -80,20 +79,17 @@ $ opam-monorepo pull
 {2 How it works}
 
 In addition to calculating the dependencies of the packages that will be
-included in the duniverse, [opam-monorepo] calculates the dependencies of the
+included in the [duniverse], [opam-monorepo] calculates the dependencies of the
 packages to be installed via [opam]. [opam-monorepo] then writes a lockfile
-that sets a variable on all dependencies it will pull, whereas [opam]-provided
-dependencies don't have variables set. In [opam] unset variables are false,
-thus [opam] will ignore the packages that [opam-monorepo] will pull.
+that sets a variable on all the dependencies it will pull, whereas [opam]-provided
+dependencies don't have variables set. In Opam, unset variables are false; thus Opam will ignore the packages that [opam-monorepo] will pull.
 
-The [opam]-provided subset of dependencies is then {e excluded} from the
-packages from the duniverse.
+The [opam]-provided subset of dependencies is then {e excluded} from the [duniverse] packages.
 
 {3 Resolution of dependency overlaps}
 
-There are cases where a dependency package appears multiple times in the
-dependency tree of a project. It can happen that a package is part of packages
-to be included in the duniverse as well as packages installed by [opam].
+There are cases where a dependency package appears multiple times in the project’s dependency tree. It’s possible that a package is part of the packages
+to be included in the [duniverse], as well as packages installed by [opam].
 
 In such cases there are two possibilities:
 
@@ -104,6 +100,6 @@ In such cases there are two possibilities:
 
 Both these approaches have advantages and disadvantages. [opam-monorepo]
 currently chooses #2 to maximize the amount of packages that are vendored, thus
-increasing the size of the reproducible set. Yet this does not mean that two
-different versions of the package can be installed - the set of packages in
-opam and the duniverse must be co-installable!
+increasing the size of the reproducible set. Yet this doesn’t mean that two
+different versions of the package can be installed. The set of packages in
+Opam and the [duniverse] must be co-installable!
--- a/doc/opam-provided.mld
+++ b/doc/opam-provided.mld
@@ -1,0 +1,93 @@
+{1 [opam]-provided dependencies}
+
+This section documents a feature that is useful if not all dependencies can be
+installed via [opam-monorepo]. This workflow is not meant for regular use but
+only as a way to enable to use [opam-monorepo] in cases where it wouldn't be
+otherwise possible.
+
+{2 Usecases}
+
+Sometimes it is not possible to put all your dependencies into the [duniverse],
+be it due to  your dependencies not building with [dune] or some tooling that
+should be just installed via [opam]. Such examples include:
+
+{ul
+{- TODO}
+}
+
+These users can opt-in into having some of their dependencies provided by
+[opam] instead of [opam-monorepo] pulling them into the duniverse.
+
+{2 Initial setup}
+
+With the default configuration [opam-monorepo] runs in full-duniverse mode,
+that is requiring all the dependencies of your [opam] files to be able to be
+built as part of the duniverse.
+
+As such it is only recommended that packages are installed via [opam] for which
+there is no way to install them via [dune], e.g. by using the
+{{:https://github.com/dune-universe/opam-overlays}opam-overlays} repository
+with dune ports.
+
+To let [opam-monorepo] know a package is to be installed via [opam] it needs to
+be marked as such in the opam file. For this the
+[x-opam-monorepo-opam-provided] field is used, which takes a single or a list
+of packages to be installed via [opam] instead of pulled into the duniverse.
+
+To configure [opam-monorepo] to avoid pulling in package [foo] into the
+duniverse use this syntax in your [opam] file:
+
+{[
+x-opam-monorepo-opam-provided: [foo]
+]}
+
+This can either be specified directly in the [opam] file or, if you use [dune]
+to generate [opam] files, in the [.template] file. In such case make sure to
+re-generate the [opam] file.
+
+{2 Usage}
+
+After you configured your [opam] file, you have to {e lock} the environment.
+
+{[
+$ opam monorepo lock
+]}
+
+After this succeeds, you will have a [.opam.locked] file, which contains all
+the dependencies (including transitive dependencies) of your project. The ones
+that [opam-monorepo] will pull into the duniverse are marked as [vendor].
+
+{[
+$ opam install --ignore-pin-depends --deps-only ./ --locked
+$ opam-monorepo pull
+]}
+
+{2 How it works}
+
+In addition to calculating the dependencies of the packages that will be
+included in the duniverse, [opam-monorepo] calculates the dependencies of the
+packages to be installed via [opam]. [opam-monorepo] then writes a lockfile
+that sets a variable on all dependencies it will pull, whereas [opam]-provided
+dependencies don't have variables set. In [opam] unset variables are false,
+thus [opam] will ignore the packages that [opam-monorepo] will pull.
+
+The [opam]-provided subset of dependencies is then {e excluded} from the
+packages from the duniverse.
+
+{3 Resolution of dependency overlaps}
+
+There are cases where a dependency package appears multiple times in the
+dependency tree of a project. It can happen that a package is part of packages
+to be included in the duniverse as well as packages installed by [opam].
+
+In such cases there are two possibilities:
+
+{ol
+{- Install the package in [opam], exclude it from duniverse}
+{- Install the package in [opam], include a duplicate in the duniverse}
+}
+
+Both these approaches have advantages and disadvantages. For simplicity
+[opam-monorepo] currently choses #1, since the dependent package needs to be
+installed in [opam] in any case and packages built in the duniverse can access
+[opam]-provided packages as well.
--- a/doc/opam-provided.mld
+++ b/doc/opam-provided.mld
@@ -8,7 +8,7 @@ otherwise possible.
 {2 Usecases}
 
 Sometimes it is not possible to put all your dependencies into the [duniverse],
-be it due to  your dependencies not building with [dune] or some tooling that
+be it due to your dependencies not building with [dune] or some tooling that
 should be just installed via [opam]. Such examples include:
 
 {ul
@@ -96,7 +96,8 @@ In such cases there are two possibilities:
 {- Install the package in [opam], include a duplicate in the duniverse}
 }
 
-Both these approaches have advantages and disadvantages. For simplicity
-[opam-monorepo] currently choses #1, since the dependent package needs to be
-installed in [opam] in any case and packages built in the duniverse can access
-[opam]-provided packages as well.
+Both these approaches have advantages and disadvantages. [opam-monorepo]
+currently chooses #2 to maximize the amount of packages that are vendored, thus
+increasing the size of the reproducible set. Yet this does not mean that two
+different versions of the package can be installed - the set of packages in
+opam and the duniverse must be co-installable!
--- a/doc/opam-provided.mld
+++ b/doc/opam-provided.mld
@@ -20,7 +20,6 @@ should be just installed via [opam]. Such examples include:
 {ul
 {- Packages not building with Dune}
 {- Packages that should not be vendored for legal reasons}
-{- Packages that just provide helper binaries like [utop], [ocamlformat]}
 }
 
 These users can opt-in to having some of their dependencies provided by [opam]
@@ -33,7 +32,7 @@ i.e., requiring all the dependencies of your [opam] files to be able to be
 built as part of the [duniverse].
 
 As such it is only recommended that packages are installed via [opam] for which
-there is no way to install them via [dune], e.g., by using the
+there is no way to build them with [dune], e.g., by using the
 {{:https://github.com/dune-universe/opam-overlays}opam-overlays} repository
 with Dune ports.
 
--- a/doc/opam-provided.mld
+++ b/doc/opam-provided.mld
@@ -5,6 +5,12 @@ installed via [opam-monorepo]. This workflow is not meant for regular use but
 only as a way to enable to use [opam-monorepo] in cases where it wouldn't be
 otherwise possible.
 
+We suggest minimizing its usage and prefer to use (and contribute) dune-ports
+from the [opam-overlays] repository. It's usage is safest with
+leaf-dependencies and those that do not have no dependency intersections with
+vendored packages. Due to possibly unexpected interactions this feature is only
+meant for advanced users and the solutions that it produces might be in flux.
+
 {2 Usecases}
 
 Sometimes it is not possible to put all your dependencies into the [duniverse],
--- a/doc/opam-provided.mld
+++ b/doc/opam-provided.mld
@@ -2,14 +2,14 @@
 
 This section documents a feature that’s useful when some dependencies cannot be
 installed via [opam-monorepo]. This workflow isn’t meant for regular use but
-only as a way to use [opam-monorepo] in cases where it wouldn't be
-otherwise possible.
+only as a way to use [opam-monorepo] in cases where it wouldn't be otherwise
+possible.
 
 We suggest minimizing its usage and prefer to use (and contribute) [dune-ports]
 from the [opam-overlays] repository. It's usage is safest with
 leaf-dependencies and those that don’t have dependency intersections with
-vendored packages. Due to possibly of unexpected interactions, this feature is only
-meant for advanced users, and the solutions it produces might be in flux.
+vendored packages. Due to possibly of unexpected interactions, this feature is
+only meant for advanced users, and the solutions it produces might be in flux.
 
 {2 Usecases}
 
@@ -23,8 +23,8 @@ should be just installed via [opam]. Such examples include:
 {- Packages that just provide helper binaries like [utop], [ocamlformat]}
 }
 
-These users can opt-in to having some of their dependencies provided by
-[opam] instead of [opam-monorepo] pulling them into the [duniverse].
+These users can opt-in to having some of their dependencies provided by [opam]
+instead of [opam-monorepo] pulling them into the [duniverse].
 
 {2 Initial setup}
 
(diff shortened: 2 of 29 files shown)


NathanReb:
So, I haven't looked at the code yet but there are a few things to clarify here.

The first is that to work well with opam, the variable should be used to mark the packages to pull, not the ones to opam install.
It should also work as it does today for opam files that don't have the variable at all, e.g. assume everything must be vendored in those cases.

The second is that we need to think carefuly about the solving step. Including the opam deps there will give a consistent result as you'll know everything can be co-installed but it then requires us to properly filter out any transitive deps that have leaked in the solution.
Having everything co-installable might also not be a requirement, for instance if you need to install a binary tool, the fact that its dependencies conflict wit the ones of your project is somewhat irrelevant.

I think a good goal here is to go for the minimum viable. The minimum is that the opam deps and duniverse can be built in the same switch, i.e. they are compatible ocaml and dune-wise. I think we should proceed as follow:
- Filter out the opam install packages from the deps formula (some packages such as ocaml and dune should be kept) of the targets packages before sending them to the solver
- Extract the dune and ocaml constraints from the deps formula of the package that have been filtered out above
- Run the solver with the filtered deps formula and the extra constraints extracted above. This gives us a solution for the duniverse, ocaml and dune.

At this point we can run the solver again with the fixed ocaml and dune from the above solution to get a solution for the opam install packages and merge it with the other solution so that everything that matters is locked there and one can just use the lockfile for all their dependency needs. The result of this merge might not always work for several reasons. We should probably stick to adding the opam install packages (not their deps) with the `=` constraint form the second solver run and let opam decide which version it will install for their deps. This is not a total lock but I think it's good enough for a first iteration.

I don't think there's an easy answer here so let's think in terms of workflow. What I'd like to advocate is the following:
```
opam switch create ./ --empty
opam monorepo lock
opam install --deps-only -t --locked ./
opam monorepo pull
```

Another option is to completely skip the second solver run and just re-inject the deps from the source opam files. It might make for a better first iteration by the way.

One last thing to note is that the lockfile should always contain the variables to mark the ones that opam should install or not. It's probably worth doing this fist in a separate PR by the way as this will enable the above workflow!

Leonidas-from-XIV:
@NathanReb I'm currently resolving everything marked as opam-provided with the solver and removing it from the set of dependencies. There might be cases where it would resolve differently than with a larger set but this seems to me to be also true if I were to manually go down the formulas as well, creating another kind of solver. This seemed to be a convenient setup but if there's any concerns I am happy to change the code.

NathanReb:
This is starting to look great!

I've done a bit of testing on the latest version and I identified a couple things:
1. It seems that the opam-provided packages and their dependencies are properly filtered out atm. I tested this on opam-monorepo itself, marking `uri` as `opam-provided`. `uri` has a dependency on `angstrom` and we don't have transitive dependencies on either of them. They were both filtered out correctly, this shows because they are not marked with the `vendored` variable in the lockfile. They do appear in the `duniverse-dirs` though which suggest they were wrongly included in the `Duniverse.t`. We need to make sure that's not the case. They also appear in the `pin-depends` field but I guess fixing this by filtering them out before building the `Duniverse.t` should fix both problems.
2. I then tried to add angstrom as a direct dependency, without marking it as `opam-provided` and the result was that it was still filtered out when I expected it would not. What should be done here is a bit less clear as we're explicitly going to state that dependency cones intersection between duniverse and opam-provided are not supported. We need to at least look into this to decide what should be done and document the filtering accordingly.

NathanReb:
One thing I just thought about: the opam-provided packages shouldn't be required to build with dune, we need to patch the solver so it knows to do that!

Leonidas-from-XIV:
1. I was also testing with `uri`. It's a good catch, fortunately it is quite easy to fix because there is code that filters out base packages and virtual packages, so adding a check there solved that issue.
2. This is indeed an issue, that we need to figure out what the right thing is in this case. At the moment, everything in the dependency tree of opam-provided packages is eliminated. This has the advantage that the transitive dependencies of opam-provided packages don't need to use dune. For them to be `opam-provided` they must be installed by opam anyways. Therefore pulling them into the duniverse would mean we need to duplicate them and the duplicate copy needs to be dune-buildable. Let's discuss the semantics that we wish for it.
3. That's an excellent observation that I completely missed out! I changed it so it runs the solver first for the opam-provided dependencies allowing non-dune builds and put the set of packages that are thus known to be installed by opam (by whichever means) in an exception list. This has some interaction with 2. because if we install `angstrom` again, the second copy has to build with dune.

Leonidas-from-XIV:
I've added a number of tests:

* A very simple one to show vendoring in action
* A more complicated one with transitive dependencies
* A variation of the previous one, but with the package to be opam-provided flipped, so demonstrate the implication

lib/opam_solve.ml
```diff
@@ -187,37 +221,87 @@ module Make_solver (Context : OPAM_MONOREPO_CONTEXT) :
...
+        |> OpamPackage.Name.Set.of_list |> Result.ok
+
+  type raw_calculation = { package : OpamPackage.t; vendored : bool }
 
-  let calculate_raw ~local_packages ~target_packages context =
-    let target_packages_names = OpamPackage.Name.Set.elements target_packages in
+  let pp_raw_calculation fmt { package; vendored = _ } =
+    Opam.Pp.package fmt package
+
+  let opam_provided_packages local_packages target_packages =
```
NathanReb, line 265: I don't think this belongs in the solver, this should rather be extracted from `Lock`. It should also use the `Source_opam_config` I think, it would simplify things a bit here.

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,91 @@
...
+
+Since we add a new custom stanza to the OPAM file, let's make sure we emit
+warnings when things are specified the wrong way, for example if the package to
+be ignored is not a string.
+
+  $ opam show --no-lint --raw -fx-opam-monorepo-opam-provided ./warning.opam
+  42
+
+  $ opam-monorepo lock warning 2>&1 | grep WARNING
+  opam-monorepo: [WARNING] Error parsing x-opam-monorepo-provided: String or List required, ignoring.
```
NathanReb, line 83: I think this could be unit tested as are other `Source_opam_config` fields

NathanReb: Also we should likely fail here, I can't think of a good reason why someone would want to keep an invalid opam-monorepo extension around.

NathanReb (commented):
It's good to finally have tests for this!

I think another good thing to test is the dependency collision thing, e.g. both the target package and an opam provided package depend on a same third one, whatever behaviour we choose there.

I have a few comments on the code but it's looking good!

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,91 @@
...
+  
+
+It should lock successfully.
+
+  $ opam-monorepo lock vendored > /dev/null
+
+The lockfile should thus contain the package `b` and mark it as `vendor` since
+opam-monorepo will vendor it.
+
+  $ opam show --no-lint --raw -fdepends ./vendored.opam.locked | grep "\"b\""
```
NathanReb, line 26: I think you can grep the file directly here, although I understand this is making it clear it comes from the dependencies of the lockfile, I'm wondering whether it's really necessary.

NathanReb: It made me realize that I wrote a bunch of `cat file | grep` in the tests... I'll fix those!

lib/opam_solve.ml
```diff
@@ -23,6 +23,8 @@ module type OPAM_MONOREPO_CONTEXT = sig
 
   val create :
     ?install_test_deps_for:OpamPackage.Name.Set.t ->
+    ?exempt_dune_for:OpamPackage.Name.Set.t ->
+    ?require_dune:bool ->
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Not for packages that OPAM is supposed to install. We need the solver to find a solution for this case and then is fair game not to require dune.

lib/opam.ml
```diff
@@ -149,27 +159,29 @@ module Package_summary = struct
     hashes : OpamHash.t list;
     dev_repo : string option;
     depexts : (OpamSysPkg.Set.t * OpamTypes.filter) list;
+    vendored : bool;
```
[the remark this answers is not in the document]
Leonidas-from-XIV: I would kinda argue that it becomes package metadata, since you could see it as this:

```ocaml

type version_constraint = Vendored | With_test | ...

{
...
  additional_version_constraints: version_constraint list;
...
}
```

I can move this away somewhere but interpreting this module/type as "opam-monorepo's view of what this particular OPAM package is" seems to me quite sensible.

lib/opam_solve.ml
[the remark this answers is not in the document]
NathanReb: I thought we were running the solver only once to find a coherent solution both for opam and duniverse installed packages. This seems to indicate we are not, am I missing something here?

lib/opam.ml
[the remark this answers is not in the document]
NathanReb: I think it's unrelated to opam and is completely tied to our view and the result of the solver. The fact that you have to explicitly pass the `vendored` boolean whenever you're building a package summary indicates that it is oddly located to me.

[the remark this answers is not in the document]
NathanReb: To clarify this module wraps up opam functionalities to have a simpler and more suited API for us so I do believe it's not the right place.

Package summary is used in `Duniverse` for instance, where the vendored status should not be handled because only vendored packages go there.

lib/opam_solve.ml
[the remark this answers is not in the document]
NathanReb: I looked further into this. Indeed I realize we do have to run the solver twice as that's the simplest way to make sure we can find a proper solution for opam-installed packages.

I have comments on how this is handled though but I'll post them where this second solver run is handled!

lib/opam_solve.ml
```diff
@@ -251,27 +335,47 @@ module Make_solver (Context : OPAM_MONOREPO_CONTEXT) :
...
+      | Ok opam_provided -> opam_provided
+      | Error msg ->
+          Logs.warn (fun l ->
+              l "Error parsing x-opam-monorepo-provided: %s, ignoring.\n" msg);
+          OpamPackage.Name.Set.empty
+    in
+
+    let allow_compiler_variants = depend_on_compiler_variants local_packages in
+    let* opam_provided_deps =
+      determine_dependencies opam_context
```
NathanReb, line 368: I don't think the opam-provided packages should be solved first.

The problem with this approach is that we don't have the insurance that the opam provided and duniverse packages require the same ocaml or dune versions. It also reverses things, opam provided packages deps prime over duniverse ones and that shouldn't be the case. Assembling a valid duniverse should be the priority here. Adding an opam provided package can strip out a dependency from the duniverse here.

One thing we can eventually try is a single solver run, where we dynamically add to a list of dune exempt packages. Happy to discuss this further offline!

lib/opam_solve.ml
```diff
@@ -59,43 +63,72 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
...
     let uses_dune =
       Opam.depends_on_dune ~allow_jbuilder depends
       || Opam.depends_on_dune ~allow_jbuilder depopts
     in
-    let summary = Opam.Package_summary.from_opam ~pkg opam_file in
+    let uses_dune =
+      match require_dune with
+      | false -> true
+      | true -> (
+          (* if it is exempt, we assume it uses dune *)
```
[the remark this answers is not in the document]
Leonidas-from-XIV: I've simplified the code. I hope it got a bit clearer.

test/bin/opam-provided.t/run.t
Leonidas-from-XIV: Moved it to `Source_opam_config` and I agree, failing makes sense.

What I am not so thrilled about is that `get` requires a `cwd` that will not be used so I specified a dummy `/` cwd but obviously this is not very nice at all.

lib/opam_solve.ml
Leonidas-from-XIV: As we have discussed offline I have implemented as follows:

1. Filter out all the opam-provided packages from all the formulas that we feed into the solver
2. Run the solver, and get a list of packages + versions. These are supposed to be dune-installable
3. Create a context (`opam_context`) where we packages are allowed not to use dune, inject a dummy package (`opam-monorepo-all-dependencies.1`) into it. The solver does only allow solving for `OpamPackage.Name.t` not `OpamPackage.t`, hence the dummy package which depends on everything from step 2 + opam-provided packages
4. Run the solver again, thus getting a solution with both our vendored packages at exactly the versions we wanted as well as the additional opam packages at versions that are co-installable with the rest

lib/opam_solve.ml
Leonidas-from-XIV: I thought it wouldn't but after moving stuff to use `Source_opam_config` you're right - this makes more sense in `Lock`. Only downside is another argument to the `Solver.compute` but I guess from a logical perspective it should not be part of the computation of dependencies but rather injected into it, so this is indeed better!

lib/opam_solve.mli
```diff
@@ -22,11 +22,12 @@ val calculate :
...
+  opam_provided:OpamPackage.Name.Set.t ->
   pin_depends:(OpamTypes.version * OpamFile.OPAM.t) OpamPackage.Name.Map.t ->
   ?ocaml_version:string ->
   ('context, 'diagnostics) t ->
   'context ->
-  ( Opam.Package_summary.t list,
+  ( Opam.Dependency_entry.t list,
     [> `Diagnostics of 'diagnostics | `Msg of string ] )
   result
 (** Calculates a solution for the provided local packages and their opam files
```
christinerose, line 33:
```suggestion
(** Calculates a solution for the provided local packages and their Opam files
```

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+  "dune" {= "2.9.1"}
+  "ocaml" {= "4.13.1"}
+  "ocaml-base-compiler" {= "4.13.1"}
+  "ocaml-config" {= "2"}
+  "ocaml-options-vanilla" {= "1"}
+
+`depends-on-b` is vendored (since it can) but `b` is opam-provided. Neat.
+
+Now for the reverse case, `depends-on-b` is opam-provided.
+
```
christinerose, line 86:
```suggestion
[depends-on-b] is vendored (since it can) but `b` is [opam]-provided. Neat.

Now for the reverse case, `depends-on-b` is [opam]-provided.

```

README.md
```diff
@@ -79,6 +79,17 @@ opam-repository) thanks to the `pin-depends`.
...
+```
+x-opam-monorepo-opam-provided: ["ocamlformat" "patdiff"]
+```
+
+This will exclude the packages from the list of packages opam-monorepo will
+pull, so they can be installed via `opam` manually.
+
 ### opam monorepo pull
 
 The `pull` command fetches the sources using the URLs in the lockfile. It benefits from the opam
```
christinerose, line 95:
```suggestion
The `pull` command fetches the sources using the URLs in the lockfile. It benefits from the Opam
```

doc/opam-provided.mld
```diff
@@ -0,0 +1,109 @@
...
+meant for advanced users and the solutions that it produces might be in flux.
+
+{2 Usecases}
+
+Sometimes it is not possible to put all your dependencies into the [duniverse],
+be it due to your dependencies not building with [dune] or some tooling that
+should be just installed via [opam]. Such examples include:
+
+{ul
+{- Packages not building with dune}
```
christinerose, line 21:
```suggestion
{- Packages not building with Dune}
```

doc/opam-provided.mld
```diff
@@ -0,0 +1,109 @@
...
+As a shortcut syntax if there is only one package it is possible to leave the
+list out and just specify the package itself:
+
+{[
+x-opam-monorepo-opam-provided: "foo"
+]}
+
+{2 Usage}
+
+After you configured your [opam] file, you have to {e lock} the environment.
```
christinerose, line 65:
```suggestion
After you configured your Opam file, you have to {e lock} the environment.
```

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+  $ opam show --no-lint --raw -fdepends ./reverse-transitive.opam
+  "dune" "depends-on-b"
+  $ opam show --no-lint --raw -fx-opam-monorepo-opam-provided ./reverse-transitive.opam
+  "depends-on-b"
+
+Since it is, we need to make its dependency, `b`, also opam-provided.
```
christinerose, line 92:
```suggestion
Since it is, we need to make its dependency, `b`, also [opam]-provided.
```

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+  "depends-on-b" {= "1"}
+  "dune" {= "2.9.1"}
+  "ocaml" {= "4.13.1"}
+  "ocaml-base-compiler" {= "4.13.1"}
+  "ocaml-config" {= "2"}
+  "ocaml-options-vanilla" {= "1"}
+
+Since we add a new custom stanza to the OPAM file, let's make sure we emit
+warnings when things are specified the wrong way, for example if the package to
+be ignored is not a string.
```
christinerose, line 109:
```suggestion
Since we add a new custom stanza to the Opam file, let's make sure we emit
warnings when things are specified the wrong way, e.g., if the package to
be ignored is not a string.
```

christinerose (changes requested):
I’ve taken a first pass at these files. I’ve suggested some formatting around terms, as we’re trying to be consistent across all documentation. The proper name of the package manager, Opam, should be capitalised as shown. I checked with Anil & Thomas, and they prefer Opam to OPAM or opam when referring to the package manager itself. 

Otherwise, `opam` should be wrapped in monospace when referring to the command, code, package, library, etc. I’m open to suggestions on how to differentiate between code/command and packages if you think it needs further visual clarification. 

I mostly commented on what I could see here (the changed parts), but please let me know when it’s merged, and I’ll go through the entire readme.md file for consistency and line editing. Then I’ll start working my way through all the Opam documentation for the same. 

Thanks for asking me to look over it! I hope my comments and suggestions are helpful.

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+
+  $ opam show --no-lint --raw -fx-opam-monorepo-opam-provided ./vendored.opam
+  
+
+It should lock successfully.
+
+  $ opam-monorepo lock vendored > /dev/null
+
+The lockfile should thus contain the package `b` and mark it as `vendor` since
+opam-monorepo will vendor it.
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Given this is a `.t` file and I think mostly we write Markdown in them it should probably be backticks:

```suggestion
`opam-monorepo` will vendor it.
```

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+  "base-threads" {= "base"}
+  "base-unix" {= "base"}
+  "dune" {= "2.9.1"}
+  "ocaml" {= "4.13.1"}
+  "ocaml-base-compiler" {= "4.13.1"}
+  "ocaml-config" {= "2"}
+  "ocaml-options-vanilla" {= "1"}
+
+Let's now check with the same opam file but this one adds `opam-provided`.
+Asking for the value will return that `b` will be provided by opam:
```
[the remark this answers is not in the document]
Leonidas-from-XIV: As above
```suggestion
Let's now check with the same Opam file but this one adds `opam`-provided.
Asking for the value will return that `b` will be provided by Opam:
```

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+
+  $ opam show --no-lint --raw -fx-opam-monorepo-opam-provided ./opam-provided.opam
+  "b"
+  $ opam-monorepo lock opam-provided > /dev/null
+
+We should be seing that this package is not marked as `vendor` so if we run
+`opam` on it, it will install the package `b`.
```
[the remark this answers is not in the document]
Leonidas-from-XIV:
```suggestion
We should see that this package is not marked as `vendor` so if we run
`opam` on it, it will install the package `b`.
```

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+  "dune" {= "2.9.1"}
+  "ocaml" {= "4.13.1"}
+  "ocaml-base-compiler" {= "4.13.1"}
+  "ocaml-config" {= "2"}
+  "ocaml-options-vanilla" {= "1"}
+
+What happens in the case that a package would be ok to vendor but the
+transitive dependency is opam-provided? In this case we have a package
+`transitive` that depends on a package that depends on a package that will
+depend transitively on an opam provided package.
```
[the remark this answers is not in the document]
Leonidas-from-XIV:
```suggestion
What happens in the case that a package would be ok to vendor but the
transitive dependency is `opam`-provided? In this case we have a package
`transitive` that depends on packages that will
depend transitively on an `opam`-provided package.
```

test/bin/opam-provided.t/run.t
Leonidas-from-XIV:
```suggestion
[depends-on-b] is vendored (since it can) but `b` is `opam`-provided. Neat.

Now for the reverse case, `depends-on-b` is `opam`-provided.

```

test/bin/opam-provided.t/run.t
Leonidas-from-XIV:
```suggestion
Since it is, we need to make its dependency, `b`, also `opam`-provided.
```

christinerose (commented):
I see in parts of run.t opam-provided is enclosed in ` ` rather than [ ] in other parts. Which one is preferred?

I also found one instance of "vendor" not enclosed in [ ] ... should it be?

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+This package does not set any `opam-provided` packages:
+
+  $ opam show --no-lint --raw -fx-opam-monorepo-opam-provided ./vendored.opam
+  
+
+It should lock successfully.
+
+  $ opam-monorepo lock vendored > /dev/null
+
+The lockfile should thus contain the package `b` and mark it as `vendor` since
```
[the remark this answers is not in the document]
Leonidas-from-XIV: `run.t` files are assumed to be Markdown (I guess?) and should use `` ` ``.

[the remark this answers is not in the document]
christinerose: That's good to know, as most of them were in [ ], so I followed that formatting. So you're saying things in monospace in `run.t` files should be wrapped in ``` instead of [ ]?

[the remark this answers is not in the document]
Leonidas-from-XIV: Yes, exactly. `[]` is only for "odoc"-markup documents like `.ml` files (source code documentation) and `.mld` files (free text documentation).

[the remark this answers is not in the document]
christinerose: Thank you for clarifying that. I'll go through `run.t` and fix those.

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
+We have a package that should not be installed via Opam. There might be various
+reasons we might not want that (e.g., it cannot be built with Dune), it is
+meant as a helper binary etc.
+
+We start with a minimal [opam-repository]:
```
christinerose, line 5:
```suggestion
We start with a minimal `opam-repository`:
```

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+
+  $ gen-minimal-repo
+
+We have a file that depends on package `b`.
```
christinerose, line 9:
```suggestion
We have a file that depends on package "b".
```

test/bin/opam-provided.t/run.t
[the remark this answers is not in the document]
christinerose: Putting b in "b" instead of single quotes ('b') or backslashes (`b`) to match the code version. Does that work for you?

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+  "base-threads" {= "base"}
+  "base-unix" {= "base"}
+  "dune" {= "2.9.1"}
+  "ocaml" {= "4.13.1"}
+  "ocaml-base-compiler" {= "4.13.1"}
+  "ocaml-config" {= "2"}
+  "ocaml-options-vanilla" {= "1"}
+
+Let's now check with the same Opam file but this one adds `opam`-provided.
+Asking for the value will return that `b` will be provided by Opam:
```
christinerose, line 38:
```suggestion
Let's now check with the same Opam file but this one adds `opam`-provided.
Asking for the value will return that "b" will be provided by Opam:
```

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+
+  $ opam show --no-lint --raw -fx-opam-monorepo-opam-provided ./opam-provided.opam
+  "b"
+  $ opam-monorepo lock opam-provided > /dev/null
+
+We should see that this package is not marked as `vendor` so if we run
+`opam` on it, it will install the package `b`.
```
christinerose, line 45:
```suggestion
We should see that this package is not marked as `vendor` so if we run
`opam` on it, it will install the package "b".
```

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+  "base-threads" {= "base"}
+  "base-unix" {= "base"}
+  "depends-on-b" {= "1" & vendor}
+  "dune" {= "2.9.1"}
+  "ocaml" {= "4.13.1"}
+  "ocaml-base-compiler" {= "4.13.1"}
+  "ocaml-config" {= "2"}
+  "ocaml-options-vanilla" {= "1"}
+
+[depends-on-b] is vendored (since it can) but `b` is `opam`-provided. Neat.
```
christinerose, line 83:
```suggestion
`depends-on-b` is vendored (since it can) but "b" is `opam`-provided. Neat.
```

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+
+Now for the reverse case, `depends-on-b` is `opam`-provided.
+
+  $ opam show --no-lint --raw -fdepends ./reverse-transitive.opam
+  "dune" "depends-on-b"
+  $ opam show --no-lint --raw -fx-opam-monorepo-opam-provided ./reverse-transitive.opam
+  "depends-on-b"
+
+Since it is, we need to make its dependency, `b`, also `opam`-provided.
```
christinerose, line 92:
```suggestion
Since it is, we need to make its dependency, "b", also `opam`-provided.
```

christinerose (changes requested):
Updated run.t with MD formatting rather than `odoc`

doc/opam-provided.mld
```diff
@@ -0,0 +1,105 @@
...
+{2 Usecases}
+
+Sometimes it is not possible to put all your dependencies into the [duniverse],
+be it due to your dependencies not building with [dune] or some tooling that
+should be just installed via [opam]. Such examples include:
+
+{ul
+{- Packages not building with Dune}
+{- Packages that should not be vendored for legal reasons}
+{- Packages that just provide helper binaries like [utop], [ocamlformat]}
```
NathanReb, line 23: I don't think this is a very good example as ocamlformat and possibly utop as well can be vendored seemlessly. It's probably best to replace this with concrete example we know will use this such as `ocaml-freestanding` or foreign language libraries such as tezos-rust-libs for instance.

lib/opam_solve.ml
```diff
@@ -59,43 +63,109 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
...
-              (version, Ok opam_file)
+            if
+              is_valid_candidate ~allow_jbuilder ~require_dune ~name ~version
+                opam_file
+            then (version, Ok opam_file)
             else (version, Error Non_dune))
       versions
 
-  let candidates { base_context; fixed_packages; allow_jbuilder } name =
+  let remove_opam_provided ~opam_provided versions =
```
NathanReb, line 140: Probably worth making this a no-op if `opam_provided` is empty rather than iterate over all opam files unconditionally.

TheLortex:
Hi, I am testing this PR to check that it works in the Mirage use case. So I am using the feature with the non-dune `ocaml-freestanding` package, which depends on the non-dune `solo5` package. Currently my issues is that with the following opam file:
```
depends: [
  "ocaml-freestanding" { build & >= "0.7.0" }
]
x-opam-monorepo-opam-provided: [ "ocaml-freestanding" ]
``` 

I obtain this lockfile:

```
depends: [
  "base-bigarray" {= "base"}
  "base-threads" {= "base"}
  "base-unix" {= "base"}
  "conf-libseccomp" {= "1"}
  "conf-pkg-config" {= "2"}
  "ocaml" {= "4.13.1"}
  "ocaml-base-compiler" {= "4.13.1"}
  "ocaml-config" {= "2"}
  "ocaml-freestanding" {= "0.7.0"}
  "ocaml-options-vanilla" {= "1"}
  "ocaml-src" {= "4.13.1"}
  "ocamlfind" {= "1.9.3" & vendor}
  "solo5" {= "0.7.0" & vendor}
]
```

I expected `solo5` to also be opam-provided, but it's actually locked ! 

It looks like there are some issues with the tests. as the behaviour I'm testing is the same as `./reverse-transitive.opam`: what I expect is written in the text: _
Now for the reverse case, `depends-on-b` is `opam`-provided. [...] Since it is, we need to make its dependency, "b", also `opam`-provided._ but the `opam show --no-lint --raw -fdepends ./reverse-transitive.opam.locked` result show the same issue.
 (https://github.com/Leonidas-from-XIV/opam-monorepo/blob/2be8fb08501812206e07ba20118ef4817dcf2d22/test/bin/opam-provided.t/run.t#L85)

In any case thanks for the huge work !

doc/opam-provided.mld
Leonidas-from-XIV: I've removed it for now, we can add it later if we find good examples. `tezos-rust-libs` feels a bit too obscure to add, maybe `ocaml-freestanding` indeed.

NathanReb:
@TheLortex thanks for the testing! We'll try to reproduce this and fix it as this is one of the intended packages that this feature is supposed to work with!

Leonidas-from-XIV:
@TheLortex You're right, the expected value of the test is wrong. At some point the algorithm made sure to never duplicate dependencies, I think before [659a7fd](https://github.com/ocamllabs/opam-monorepo/pull/234/commits/659a7fd56c741596f922900fb25af1f099d47352) and I adjusted it wrong. This seems to be a bug, because indeed `b` should not be marked as `vendor` as the vendored dependencies do not depend on it.

Leonidas-from-XIV:
@TheLortex Can you take another look? There was indeed an error and I ended up just adding debug/info prints in relevant places to be able to see why a package is classified as either (so `--verbosity=debug` might show some interesting insights why things are wrong).

test/bin/opam-provided.t/run.t
```diff
@@ -0,0 +1,125 @@
...
+
+Since we add a new custom stanza to the Opam file, let's make sure we emit
+warnings when things are specified the wrong way, e.g., if the package to
+be ignored is not a string.
+
+  $ opam show --no-lint --raw -fx-opam-monorepo-opam-provided ./warning.opam
+  42
+
+  $ opam-monorepo lock warning > /dev/null
+  opam-monorepo: String or List required
```
NathanReb, line 115: Shouldn't the location in the file be printed here?

Leonidas-from-XIV: That's a a good point. I looked how the other code does this and adjusted the code to match. Quite neat functionality.

NathanReb (approved):
Looks good! This is long overdue!

TheLortex:
I'm still having issues using the feature. See this opam file:
```
opam-version: "2.0"
name: "noop"
maintainer: "dummy"
authors: "dummy"
homepage: "dummy"
bug-reports: "dummy"
dev-repo: "git://dummy"
synopsis: "Unikernel noop - monorepo dependencies"

depends: [
  "ocaml-freestanding" { build & >= "0.7.0" }
]
x-opam-monorepo-opam-provided: [ "ocaml-freestanding" ]
```

Opam-monorepo crashes with the following error:

```
               (Failure "duplicate entry ocaml")
               Called from OpamStd.Set.Make.fold.(fun) in file "src/core/opamStd.ml", line 190, characters 15-23
               Called from Stdlib__Set.Make.iter in file "set.ml", line 378, characters 35-38
               Called from Stdlib__Set.Make.iter in file "set.ml", line 378, characters 25-33
               Called from OpamStd.Set.Make.fold in file "src/core/opamStd.ml", line 189, characters 6-62
               Called from Duniverse_lib__Opam_solve.Make_solver.build_context in file "lib/opam_solve.ml", line 285, characters 6-69
               Called from Duniverse_lib__Opam_solve.Make_solver.calculate in file "lib/opam_solve.ml", line 505, characters 12-187
               Called from Duniverse_lib__Opam_solve.calculate in file "lib/opam_solve.ml" (inlined), line 642, characters 2-134
               Called from Duniverse_cli__Lock.calculate_opam.(fun) in file "cli/lock.ml", line 302, characters 14-192
               Called from OpamSwitchState.with_.(fun) in file "src/state/opamSwitchState.ml", line 1211, characters 14-18
               Re-raised at OpamStd.Exn.finalise in file "src/core/opamStd.ml", line 1372, characters 4-38
               Called from OpamStd.Exn.finally in file "src/core/opamStd.ml", line 1375, characters 10-14
               Re-raised at OpamStd.Exn.finalise in file "src/core/opamStd.ml", line 1372, characters 4-38
               Called from OpamGlobalState.with_ in file "src/state/opamGlobalState.ml", line 186, characters 14-18
               Re-raised at OpamStd.Exn.finalise in file "src/core/opamStd.ml", line 1372, characters 4-38
               Called from Duniverse_cli__Lock.run in file "cli/lock.ml", line 449, characters 4-143
               Called from Cmdliner_term.app.(fun) in file "src/cmdliner_term.ml", line 24, characters 19-24
               Called from Cmdliner_term.term_result.(fun) in file "src/cmdliner_term.ml", line 40, characters 25-32
               Called from Cmdliner_eval.run_parser in file "src/cmdliner_eval.ml", line 34, characters 37-44
```

Leonidas-from-XIV:
@TheLortex Thanks for the testing, I appreciate it. I tried with your example first and it didn't work because the solver picked 4.14 which is not compatible with `ocaml-freestanding`, so I realized you're using `--ocaml-version` to lock it.

We did indeed have an issue there, because in certain cases we inject the compiler constraint from the command line twice, but the code was written in a way where this would crash. I made sure for it to work as long as the constraint is unchanged.

@NathanReb  I added a small regression test into the cram test (that fails with the exact error @TheLortex describes) so this should not happen in the future. It is a bit more code than absolutely necessary because I wanted to make sure the constraint is equal to the constraint that already exists - otherwise there would be some tricky solver problem that we should not just ignore.

hannesm:
...impatiently waiting for this to be merged and released... thanks for all your hard work on it :)

NathanReb:
@hannesm it should get in today. We will also focus on the `0.3` release in the next few days so this should all soon be available!

NathanReb (approved):

hannesm:
Thanks again, for the work on this, and for merging it. I'm wondering @NathanReb 8 days ago wrote "release in the next few days", is there still a release planned, or have there been show stoppers encountered? Thanks for you time.

Leonidas-from-XIV:
@hannesm Yes, we are planning a 0.3.0 release soon. There is a milestone with the missing things for 0.3.0, the only potential show stopper is about the opam `vendor` variables that needs to be investigated beforehand.

PR #239, opened 2021-12-13
Move Extra_field out of Lockfile

NathanReb:
This is strictly a refactoring and a small one but it was necessary for the work I am currently doing on reproducible `lock` and thought it might also be of use for #234!

This should make rebasing either of these features easier along with reducing the size of the PRs so it seemed worthwile to extract it to its own PR.

### DIFF
--- a/lib/opam.mli
+++ b/lib/opam.mli
@@ -51,6 +51,35 @@ module Pp : sig
   val url : OpamUrl.t Fmt.t
 end
 
+module Extra_field : sig
+  (** Module for parsing and printing of opam file extensions
+      specific to opam-monorepo *)
+
+  type 'a t
+  (** Type of extra opam field holding values of type ['a]. *)
+
+  val make :
+    name:string ->
+    to_opam_value:('a -> OpamParserTypes.FullPos.value) ->
+    from_opam_value:
+      (OpamParserTypes.FullPos.value -> ('a, [ `Msg of string ]) result) ->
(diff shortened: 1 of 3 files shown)


NathanReb:
As a follow up to this in my PR, I'll add a `Source_opam_file` module to regroup all utility functions for target packages opam files, including parsing extra fields, merging them etc... but this doesn't make sense on its own so I did not include it.

lib/opam.mli
```diff
@@ -51,6 +51,35 @@ module Pp : sig
...
+  val name : _ t -> string
+  (** Return the full name of the field *)
+
+  val set : 'a t -> 'a -> OpamFile.OPAM.t -> OpamFile.OPAM.t
+  (** Sets the field in the given opam file, potentially overwriting the
+      previous value. *)
+
+  val get : 'a t -> OpamFile.OPAM.t -> ('a, [ `Msg of string ]) result option
+  (** Returns the optional value of the given extra field in the given opam
+      file. *)
```
Leonidas-from-XIV, line 80: I would reformulate it a bit to be clearer, something like

```suggestion
  (** Returns the value of the given extra field in the given opam file if
      the extra field is set. *)
```

Leonidas-from-XIV (approved):
Yes, this refactoring makes sense to me. It is nice that `add` became `set` which is in my opinion a bit more reasonable.

PR #245, opened 2022-01-13
Only find candidates that can't be built if there is no solution

Leonidas-from-XIV:
We found this issue while pairing.

We should only look at rejected solutions when there is no selected solution. In a scenario where there is 1 selected candidate that builds with dune and all others don't build with dune this would produce a wrong error.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,11 @@
 
 ### Fixed
 
+- Only show an error that there are no packages that build with dune when there
+  is no solution. The code would otherwise report an that no packages build
+  with dune if the selected package builds with dune, leading to confusion.
+  (#245, @Leonidas-from-XIV)
+
 ### Removed
 
 ### Security
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,10 +13,8 @@
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -13,6 +13,11 @@
 
 ### Fixed
 
+- Only show an error that there are no packages that build with dune when there
+  is no solution. The code would otherwise report an that no packages build
+  with dune if the selected package builds with dune, leading to confusion.
+  (#245, @Leonidas-from-XIV)
```
NathanReb, line 19: What would you think of this phrasing instead:
```suggestion
- Fix a bug where a package which had a single version that built with dune and got selected by the solver
  would be reported has having no version building with dune. (#245, @Leonidas-from-XIV)
```
I think it make it a bit clearer what the bug was.

NathanReb (commented):
Looks good, thanks for fixing this!

I just made a suggestion to improve the changelog entry but this is otherwise good to go!

Leonidas-from-XIV:
Yes, I think it is better. Writing changelogs from a user perspective is quite hard when you look too much at the code. Thanks for the improvement!

PR #246, opened 2022-01-13
Factor opam value parsing out of Lockfile

NathanReb:
This allows me to use this code for the explicit repo feature!

### DIFF
--- a/lib/lockfile.ml
+++ b/lib/lockfile.ml
@@ -208,8 +208,8 @@ module Duniverse_dirs = struct
           in
           Ok (OpamUrl.of_string url, (dir, hashes))
       | _ ->
-          Opam.Pos.value_errorf ~value
-            "Expected a list [ \"url\" \"repo name\" [<hashes>] ]"
+          Opam.Pos.unexpected_value_error
+            ~expected:"a list [ \"url\" \"repo name\" [<hashes>] ]" value
     in
     let* bindings = Opam.Value.List.from_value ~elm_from_value value in
     Ok (OpamUrl.Map.of_list bindings)
--- a/lib/lockfile.ml
+++ b/lib/lockfile.ml
@@ -49,13 +49,12 @@ module Version = struct
               your opam-monorepo plugin."
              pp t))
 
-  let to_opam_value t =
-    Opam.Pos.with_default (OpamParserTypes.FullPos.String (to_string t))
+  let to_opam_value t = Opam.Value.String.to_value (to_string t)
 
   let from_opam_value value =
-    match (value : OpamParserTypes.FullPos.value) with
-    | { pelem = String s; _ } -> from_string s
-    | _ -> Opam.Pos.value_errorf ~value "Expected a string"
+    let open Result.O in
+    let* str = Opam.Value.String.from_value value in
+    from_string str
 
   let field = Extra_field.make ~name:"version" ~to_opam_value ~from_opam_value
 end
@@ -64,31 +63,21 @@ module Root_packages = struct
   type t = OpamPackage.Name.Set.t
 
   let to_opam_value t =
-    let open OpamParserTypes.FullPos in
     let sorted =
       t |> OpamPackage.Name.Set.elements
       |> List.map ~f:OpamPackage.Name.to_string
       |> List.sort ~cmp:String.compare
     in
-    let pelem =
-      List
-        (Opam.Pos.with_default
-           (List.map ~f:(fun s -> Opam.Pos.with_default (String s)) sorted))
-    in
-    Opam.Pos.with_default pelem
+    Opam.Value.List.to_value ~elm_to_value:Opam.Value.String.to_value sorted
 
   let from_opam_value value =
     let open Result.O in
-    let open OpamParserTypes.FullPos in
-    let elm_from_value { pelem; _ } =
-      match pelem with
-      | String s -> Ok (OpamPackage.Name.of_string s)
-      | _ -> Opam.Pos.value_errorf ~value "Expected a string"
+    let elm_from_value value =
+      let+ str = Opam.Value.String.from_value value in
+      OpamPackage.Name.of_string str
     in
-    match value.pelem with
-    | List { pelem; _ } ->
-        Result.List.map ~f:elm_from_value pelem >>| OpamPackage.Name.Set.of_list
-    | _ -> Opam.Pos.value_errorf ~value "Expected a list"
+    Opam.Value.List.from_value ~elm_from_value value
+    >>| OpamPackage.Name.Set.of_list
 
   let field =
     Extra_field.make ~name:"root-packages" ~to_opam_value ~from_opam_value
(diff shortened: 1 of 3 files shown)


lib/opam.ml
```diff
@@ -238,3 +238,29 @@ module Pos = struct
     let pos = from_value value in
     errorf ~pos fmt
 end
+
+module Value = struct
+  module String = struct
+    let from_value value =
+      match (value : OpamParserTypes.FullPos.value) with
```
Leonidas-from-XIV, line 245: Maybe this would be better to attach to the line above? Or as part of the qualification of the record label?

lib/opam.ml
```diff
@@ -238,3 +238,29 @@ module Pos = struct
...
+      | { pelem = String s; _ } -> Ok s
+      | { pos; _ } -> Pos.errorf ~pos "Expected a string"
+
+    let to_value s = Pos.with_default (OpamParserTypes.FullPos.String s)
+  end
+
+  module List = struct
+    let from_value ~elm_from_value value =
+      match (value : OpamParserTypes.FullPos.value) with
```
Leonidas-from-XIV, line 254: Similar as above.

lib/opam.ml
```diff
@@ -238,3 +238,29 @@ module Pos = struct
...
+    let from_value value =
+      match (value : OpamParserTypes.FullPos.value) with
+      | { pelem = String s; _ } -> Ok s
+      | { pos; _ } -> Pos.errorf ~pos "Expected a string"
+
+    let to_value s = Pos.with_default (OpamParserTypes.FullPos.String s)
+  end
+
+  module List = struct
+    let from_value ~elm_from_value value =
```
Leonidas-from-XIV, line 253: Maybe calling it `element` or `item` could be a better naming? I don't know, there isn't a good convention I am aware of, it's just that `elm_from_value` sounds oddly complicated for what is essentially `~f` in `List.map`. Maybe `~f` could actually be a good alternative since it takes the same `'a` as `from_value` takes a `'a list` of?

lib/opam.ml
```diff
@@ -238,3 +238,29 @@ module Pos = struct
     let pos = from_value value in
     errorf ~pos fmt
 end
+
+module Value = struct
+  module String = struct
+    let from_value value =
+      match (value : OpamParserTypes.FullPos.value) with
+      | { pelem = String s; _ } -> Ok s
+      | { pos; _ } -> Pos.errorf ~pos "Expected a string"
```
Leonidas-from-XIV, line 247: Given this function is now reusable, maybe the error message could specify what it got instead of the string? That could make it easier for users who run into this to catch which part of the OPAM file caused the error.

Something like

```ocaml
| { pelem; pos; } -> Pos.errorf ~pos "Expected a string, got %a" pp_pelem pelem
```

so that `pp_pelem` could also be used in `Opam.Value.from_value`?

lib/lockfile.ml
```diff
@@ -195,74 +184,51 @@ module Duniverse_dirs = struct
...
-       pelem =
-         List
-           {
-             pelem = [ { pelem = String url; _ }; { pelem = String dir; _ } ];
-             _;
-           };
-       _;
-      } ->
+    let elm_from_value value =
+      let* l = Opam.Value.List.from_value ~elm_from_value:Rresult.R.ok value in
```
Leonidas-from-XIV, line 200: Since we target 4.08+ this should work:

```suggestion
      let* l = Opam.Value.List.from_value ~elm_from_value:Result.ok value in
```

Leonidas-from-XIV (commented):
This PR makes the parsing a lot clearer, it's great!

Now that it happens in one place, maybe it would make sense to make the error messages a bit better since the code is reusable?

lib/opam.ml
NathanReb: Note that the error is reported along with the location so I'm not sure there is a lot of value in describing what is there.

We could further improve this error reporting to pretty print the surrounding parts of the file, a la OCaml compiler but this feels a bit overkill for now I think.

To calrify what I have in mid would be something like:
```
In opam file foo.opam, line 7, char 5 - 23: Expected a string
...
6-    "A proper string
7-    [ "A String wrapped in a list by mistake"]
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
```

If you're okay with this, I suggest we keep this PR a refactoring and in the future if we receive complaints with error message in opam files extensions parsing we can eventually put some effort into improving them. If you do feel like the current error messages are specifically bad, we can eventually improve it in a subsequent PR as suggested above.

lib/lockfile.ml
NathanReb: Ah good catch!

lib/opam.ml
NathanReb: I guess this is mostly a matter of personal preferences I tend to do this to avoid having to disambiguate constructors or record fields in the first pattern (which I personally find kind of ugly looking) or having to use a local open. This makes it clear that we simply want to pattern match over the concrete type.

We could move it to the function argument if you prefer, I don't particularly mind. I just find it quite nice to do it here as it is possible to consistently annotate in the pattern match rather than doing it elsewhere since we're not always pattern matching over one of the argument.

lib/opam.ml
NathanReb: I think `f` is a bit to generic here considering we need a very specific kind of function in that case, as opposed to `List.map`. I believe having a good argument name will make it easier to read code using those combinator-like functions.

I'm happy to change it to `element` as `elm` might be less clear. I went for `elm` as it's used in `Set.S` from the standard library and is slightly shorter than `element` which is fairly consistently used across the standard library documentation for the `List` module.

Leonidas-from-XIV: Maybe just `~elm` in such case, to match the Stdlib?

lib/opam.ml
Leonidas-from-XIV: I agree that doing a proper error reporting with syntax is way beyond the scope of this PR: this is complex stuff and I would even argue, not particularly necessary here. Lots of work for little reward.

I still think that that printing what the program sees here would be quite easy to do. I find it useful e.g. when the OCaml compiler tells me what type it expected vs what type it got. But admittedly this can also easily done later so I'd say it is up to you, I don't consider this a must-have feature for this PR :-)

lib/opam.ml
NathanReb: I'm a bit confused now, I don't really understand why `elm` would be an improvement over what we currently have.
Could you explain what confused or bothered you with `elm_from_value` in the first place? I think this would help me fix this naming problem.

I can try better explaining why I named it like this as well:
To clarify, in the std lib, `elm` is the type name for the elements of a `Set.t`, and the term `element` is used in the documentation of list to describe elements of a list. It's never used as a labelled argument directly or at least not that I know of.

Here we expect a function for this argument, a function that converts opam values. This function will be applied to all elements of a list to convert an opam value list into an ocaml list, hence my original naming. I would find `elm` very confusing as it doesn't help me understand what should be passed here. Again, the interface will likely clarify that but I think it's a good property to make code using those functions readable, without having to jump to the doc or using merlin to show the type of the arguments.

Leonidas-from-XIV: I think my confusion is that the argument `elm_from_value` when I first saw it is a rather complex name where I expect something complicated to happen, like some very specific functionality. Then then reading continuing to read the function I realized that the reality is much simpler, this is just the recursive application (as is basically apparent from the signature itself). This makes it in my mind more comparable to `List.map` which applies some named/`f`-named function to its elements or even more comparable, `Fmt.list` which takes an `'a Fmt.t` as an argument where it is pretty clear what it will be used for.

As in a hypothetical alternative
```ocaml
Fmt.list : ?sep:unit Fmt.t -> ~elm_fmt:'a Fmt.t -> 'a list Fmt.t
```
I don't think the signature would any more understandable by giving the argument that gets applied on every element a complex-sounding name.

I don't want to bikeshed this, so if you prefer `elm_from_value` that's fine with me. This is just my explanation why I think a simple name can be good enough even if it doesn't exactly name what it is.

lib/opam.ml
NathanReb: Added just that!

lib/opam.ml
NathanReb: ok, I removed the label alltogether then!

Leonidas-from-XIV (approved):

PR #248, opened 2022-01-18
Better missing version error

### LINKED ISSUE #215: Incorrect error message if attempting to lock a version that doesn't exist
When the `opam` file requires a package that doesn't exist (e.g. due to stale `opam-repository` metadata locally or just a typo) the error message will not contain the real cause, instead it will complain that there is no version that builds with dune:

To reproduce: Do not add the dune-port overlay (hence no version of `fmt` is buildable) and depend on (at time of writing non-existing) version `>= 1.0.0`. The solver will correctly reject all versions of fmt but not mention that there doesn't exist a `fmt` package which would pass the version constraint.

Leonidas-from-XIV:
I've continued to play around with what info can be retrieved from the map and with a bit of squinting (extracting the version restriction from another rejection case) I was able to test the version check on the packages that we exclude as not building with dune.

If I specify `fmt {>= "1.0.0"}` in my `opam` file I get this error message now:

```
opam-monorepo: [ERROR] Could not find any package that would satisfy the version constrants: fmt
                       >= 1.0.0
```

It is somewhat involved and hairy so I am not entirely sure whether this is worth it so I'm opening this PR as a draft, more of an RFC to see if we even want to do stuff this way or whether we should patch the 0install solver to expose better information instead.

Closes #215

### DIFF
--- a/cli/lock.ml
+++ b/cli/lock.ml
@@ -216,7 +216,7 @@ let could_not_determine_version offending_packages =
 let interpret_solver_error ~repositories solver = function
   | `Msg _ as err -> err
   | `Diagnostics d ->
-      (match Opam_solve.no_matching_versions solver d with
+      (match Opam_solve.unavailable_versions_due_to_constraints solver d with
       | [] -> ()
       | offending_packages -> could_not_determine_version offending_packages);
       (match Opam_solve.not_buildable_with_dune solver d with
--- a/cli/lock.ml
+++ b/cli/lock.ml
@@ -212,8 +212,7 @@ let could_not_determine_version offending_packages =
   let pp_offending_packages = Fmt.list ~sep:Fmt.comma pp_offending_package in
   Logs.err (fun l ->
       l
-        "There is no eligible package that matches %a. Make sure a dune port \
-         exists."
+        "There is no eligible package that matches %a."
         pp_offending_packages offending_packages)
 
 let interpret_solver_error ~repositories solver = function
--- a/cli/lock.ml
+++ b/cli/lock.ml
@@ -211,9 +211,8 @@ let could_not_determine_version offending_packages =
   in
   let pp_offending_packages = Fmt.list ~sep:Fmt.comma pp_offending_package in
   Logs.err (fun l ->
-      l
-        "There is no eligible package that matches %a."
-        pp_offending_packages offending_packages)
+      l "There is no eligible package that matches %a." pp_offending_packages
+        offending_packages)
 
 let interpret_solver_error ~repositories solver = function
   | `Msg _ as err -> err
--- a/cli/lock.ml
+++ b/cli/lock.ml
@@ -206,13 +206,12 @@ let could_not_determine_version offending_packages =
   in
   let s = OpamFormula.string_of_formula f in
   let pp_version_formula = Fmt.using s Fmt.string in
-  let pp_offending_package =
-    Fmt.pair ~sep:Fmt.sp Opam.Pp.package_name pp_version_formula
-  in
-  let pp_offending_packages = Fmt.list ~sep:Fmt.comma pp_offending_package in
-  Logs.err (fun l ->
-      l "There is no eligible package that matches %a." pp_offending_packages
-        offending_packages)
+  List.iter
+    ~f:(fun (name, formula) ->
+      Logs.err (fun l ->
+          l "There is no eligible version of %a that matches %a"
+            Opam.Pp.package_name name pp_version_formula formula))
+    offending_packages
 
 let interpret_solver_error ~repositories solver = function
   | `Msg _ as err -> err
--- a/cli/lock.ml
+++ b/cli/lock.ml
@@ -212,8 +212,8 @@ let could_not_determine_version offending_packages =
   let pp_offending_packages = Fmt.list ~sep:Fmt.comma pp_offending_package in
   Logs.err (fun l ->
       l
-        "Could not find any package that would satisfy the version constrants: \
-         %a"
+        "There is no eligible package that matches %a. Make sure a dune port \
+         exists."
         pp_offending_packages offending_packages)
 
 let interpret_solver_error ~repositories solver = function
--- a/cli/lock.ml
+++ b/cli/lock.ml
@@ -192,9 +192,36 @@ let display_verbose_diagnostics = function
   | None -> false
   | Some l -> l >= Logs.Info
 
+let could_not_determine_version offending_packages =
+  let f (relop, version) =
+    let pp_relop fmt = function
+      | `Eq -> Fmt.pf fmt "="
+      | `Geq -> Fmt.pf fmt ">="
+      | `Gt -> Fmt.pf fmt ">"
+      | `Leq -> Fmt.pf fmt "<="
+      | `Lt -> Fmt.pf fmt "<"
+      | `Neq -> Fmt.pf fmt "!="
+    in
+    Fmt.str "%a %a" pp_relop relop Opam.Pp.version version
+  in
+  let s = OpamFormula.string_of_formula f in
+  let pp_version_formula = Fmt.using s Fmt.string in
+  let pp_offending_package =
+    Fmt.pair ~sep:Fmt.sp Opam.Pp.package_name pp_version_formula
+  in
+  let pp_offending_packages = Fmt.list ~sep:Fmt.comma pp_offending_package in
+  Logs.err (fun l ->
+      l
+        "Could not find any package that would satisfy the version constrants: \
+         %a"
+        pp_offending_packages offending_packages)
+
 let interpret_solver_error ~repositories solver = function
   | `Msg _ as err -> err
   | `Diagnostics d ->
+      (match Opam_solve.no_matching_versions solver d with
+      | [] -> ()
+      | offending_packages -> could_not_determine_version offending_packages);
       (match Opam_solve.not_buildable_with_dune solver d with
       | [] -> ()
       | offending_packages ->
--- a/lib/opam_solve.ml
+++ b/lib/opam_solve.ml
@@ -271,46 +271,51 @@ module Make_solver (Context : OPAM_MONOREPO_CONTEXT) :
     let _, version_restriction = Solver.formula restriction in
     Some version_restriction
 
+  (* [true] if the rejected package would've satisified the version constraint if it
+     wouldn't have been a [Model_rejection] *)
+  let model_rejection_of_eligible_version version_restriction (model, reason) =
+    match reason with
+    | `Model_rejection _ -> (
+        match Solver.version model with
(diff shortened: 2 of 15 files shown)


Leonidas-from-XIV:
With @NathanReb's very valuable hint that I can get the rejection condition from `Note.t` (even including the package that required the rejection) I feel this code is way more sensible and think it is in better shape to be merged.

NathanReb:
Do you think we could extract a test for this? That would also help seeing what the output looks like!

Leonidas-from-XIV:
Yes, that's definitely the idea! Which is why I rebased on `main`. It would be nice to get #254 in too, so I can use the "nicer" way of testing the repos directly, since it is practically ready to be merged.

test/bin/invalid-package-version.t/run.t
```diff
@@ -0,0 +1,46 @@
...
+
+(grep appends a NUL byte at the end, hence the head call, this is not important
+to the test)
+
+  $ opam-monorepo lock toonew 2> errors
+  ==> Using 1 locally scanned package as the target.
+  [1]
+  $ grep -Pazo "(?s)opam-monorepo: \[ERROR\].*(?=opam-monorepo)" < errors | head --bytes=-1
+  opam-monorepo: [ERROR] There is no eligible package that matches a
+                         >= 1.0. Make sure a dune port exists.
```
NathanReb, line 46: We should try to make sure the constraint is properly displayed here. It's cut between the two lines which is not great.

I don't think `Make sure a dune port exists.` should be part of this error as it is unrelated. If no regular version matches the constraint, there is no reason why a dune port would.

lib/opam_solve.ml
```diff
@@ -251,6 +254,65 @@ module Make_solver (Context : OPAM_MONOREPO_CONTEXT) :
...
+                  | _ -> None)
+                notes
+              |> Option.map ~f:Solver.formula
+            in
+            let rejects, _reason =
+              Solver.Diagnostics.Component.rejects component
+            in
+            (* find only the model rejections that is those that we have rejected as e.g. not building with dune *)
+            let model_rejected =
+              List.filter_map
```
NathanReb, line 285: I'm not sure it's worth filtering those out first, I think we can do the check in a single pass.

NathanReb (commented):
This looks good, thanks!

I have a few comments code-wise.

I also think we should add a regression test for the original issue, the one with no dune compatible versions as well. That is, if for a package, no version build with dune and no version match the constraint, only the latter is reported.

NathanReb:
Also looking at the output, I think we should consider dropping the base diagnosis if we find such packages or packages that don't build with dune. We can do this in a separate PR but I'm curious to know what you think about such an approach!

cli/lock.ml
```diff
@@ -192,9 +192,34 @@ let display_verbose_diagnostics = function
...
+    Fmt.str "%a %a" pp_relop relop Opam.Pp.version version
+  in
+  let s = OpamFormula.string_of_formula f in
+  let pp_version_formula = Fmt.using s Fmt.string in
+  let pp_offending_package =
+    Fmt.pair ~sep:Fmt.sp Opam.Pp.package_name pp_version_formula
+  in
+  let pp_offending_packages = Fmt.list ~sep:Fmt.comma pp_offending_package in
+  Logs.err (fun l ->
+      l "There is no eligible package that matches %a." pp_offending_packages
```
NathanReb, line 214: I wonder if this wouldn't be more accurate and clear if it said "There is no eligible version of <pkgname> that matches <constraint>".
We would have to print this for each offending package but I think it would make the whole thing much clearer.

With the current situation if there are several offending packages, it would read like this:
```
There is no eligible package that matches a >= 1.0, b < 0.5
```
and it's not entirely clear whether it's one package for each or one for all from the phrasing which I think can make things confusing to users.

lib/opam_solve.ml
```diff
@@ -251,6 +254,65 @@ module Make_solver (Context : OPAM_MONOREPO_CONTEXT) :
       rolemap []
     |> List.filter_map ~f:Solver.package_name
 
+  let find_version_restriction component =
+    let open Option.O in
+    let notes = Solver.Diagnostics.Component.notes component in
+    let* restriction =
+      List.find_map
```
NathanReb, line 261: I'm not sure that's how we should do it here. I think we should rather merge the different restrictions we find together. For instance the solver could fail to find a solution because some package depend on `a > 1.2` and another on `a < 2.0`.
There could be versions that match either of those, for instance `a.1.2` and `a.2.0` but none that match both.

NathanReb (commented):
I did another quick pass as I thought you already made some changes. I think I was a little early to the party but I think I found some nice improvements we could make nonetheless!

Let me know what you think!

cli/lock.ml
Leonidas-from-XIV: This is indeed much better. I think it reveals that I only ever tested it with one package that doesn't match so the wording was all wrong, while the code (and reality) supports more than one thing being wrong. Fixed this now, I it is much better.

lib/opam_solve.ml
Leonidas-from-XIV: Instead of `find_map` I could `filter_map` and `&&` all the constraints together (and potentially deduplicate them before, since I assume `a > 1.2` and `a < 2.0` would appear multiple times). Does that sound sensible as a solution to you?

This would be a good case of writing a test case first and implementing the solution in a second step.

NathanReb: Yeah that sounds great!

I expect proper deduplication might require a bit of work to get right. For instance, `a < 2.2` would be redundant with `a < 2.0`. In such a case we'd prefer to only display the latter. I think it's fine if we iteratively improve this, I think users would prefer a correct but redundant output to an incorrect but concise one!

lib/opam_solve.ml
```diff
@@ -251,6 +254,65 @@ module Make_solver (Context : OPAM_MONOREPO_CONTEXT) :
       rolemap []
     |> List.filter_map ~f:Solver.package_name
 
+  let no_matching_versions diagnostics =
```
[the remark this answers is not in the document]
Leonidas-from-XIV: I have renamed it into `unavailable_versions_due_to_constraints`. Not a great name either but hopefully better. I'm open to other suggestions!

lib/opam_solve.ml
```diff
@@ -251,6 +254,65 @@ module Make_solver (Context : OPAM_MONOREPO_CONTEXT) :
...
+    let rolemap = Solver.diagnostics_rolemap diagnostics in
+    Pkg_map.fold
+      (fun pkg component acc ->
```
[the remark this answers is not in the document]
Leonidas-from-XIV: I would prefer to keep the anonymous function for the `fold` (since it is quite specific that it does it is better to see that the arguments and their types are meant to align with the signature for `fold`) but I've tried to factor out more verbose internal helpers into named functions and give them somewhat understandable names.

lib/opam_solve.ml
NathanReb: Nah I think it's fine, we can refine it later, this is already helpful in its current state!

lib/opam_solve.ml
```diff
@@ -251,6 +254,84 @@ module Make_solver (Context : OPAM_MONOREPO_CONTEXT) :
...
     |> List.filter_map ~f:Solver.package_name
 
+  let determine_version_restriction component =
+    let open Option.O in
+    let restrictions =
+      component |> Solver.Diagnostics.Component.notes
+      |> List.map ~f:(function
+           | Solver.Diagnostics.Note.UserRequested restriction ->
+               [ restriction ]
+           | Restricts (_other_role, _impl, restrictions) -> restrictions
```
NathanReb, line 264: At some point I think we should collect `other_role` here as it would help improving the error message!

lib/opam_solve.ml
```diff
@@ -251,6 +254,84 @@ module Make_solver (Context : OPAM_MONOREPO_CONTEXT) :
...
+                ~f:(model_rejection_of_eligible_version version_restriction)
+                rejects
+            in
+            (* if it is unavailable, construct info on why *)
+            let* unavailable_pkg =
+              match model_rejections_would_match_version with
+              | false -> None
+              | true -> Some (pkg_name, version_restriction)
+            in
+            unavailable_pkg :: unavailable)
```
NathanReb, line 332: The use of the option and the `let*` here is a bit confusing I think, especially since this is the returning block in a way.
What would you think about turning this into:
```suggestion
            match model_rejections_would_match_version with
            | false -> unavailable
            | true -> (pkg_name, version_restriction)::unavailable)
```

test/bin/invalid-package-version.t/run.t
```diff
@@ -0,0 +1,55 @@
...
+
+opam-monorepo solver should successfully pick a.0.1:
+
+  $ opam-monorepo lock existing
+  ==> Using 1 locally scanned package as the target.
+  ==> Found 9 opam dependencies for the target package.
+  ==> Querying opam database for their metadata and Dune compatibility.
+  ==> Calculating exact pins for each of them.
+  ==> Wrote lockfile with 1 entries to $TESTCASE_ROOT/existing.opam.locked. You can now run opam monorepo pull to fetch their sources.
+  $ cat existing.opam.locked | grep "\"a\"\s\+{"
```
NathanReb, line 26: That's on me but this can be turned into:
```suggestion
  $ grep "\"a\"\s\+{" existing.opam.locked 
```

test/bin/invalid-package-version.t/run.t
```diff
@@ -0,0 +1,55 @@
...
+  $ opam-monorepo lock toonew 2> errors
+  ==> Using 1 locally scanned package as the target.
+  [1]
+  $ grep -Pazo "(?s)opam-monorepo: \[ERROR\].*(?=opam-monorepo)" < errors | head --bytes=-1
+  opam-monorepo: [ERROR] There is no eligible version of a that matches >= 1.0
+
+We should also produce the right error message with all the constraints when we have multiple constaints
+
+  $ opam show --no-lint --raw -fdepends ./multiple-constraint.opam
+  "dune" "a" {>= "1.1" & < "2.0"}
```
NathanReb, line 50: I think this won't be testing the right thing. To test the initial problem, the constraint have to come from different packages I think. This can be done by introducing another package, itself depending on a!

NathanReb (commented):
Looks much better, thanks! I have a few minor comments but once they're fixed this is good to go!

test/bin/invalid-package-version.t/run.t
NathanReb: After offline discussions we concluded that there's no easy way to test this scenario and we can improve this part of the testing as we start working on the generic solver error messages.

NathanReb (approved):

PR #252, opened 2022-02-02
Add debug log to list locally detected packages

NathanReb:
This can help understand the baseline for the solver when debugging. It will also prove helpful for users support!


Leonidas-from-XIV (approved):
Yeah, I can see that coming in useful!

PR #253, opened 2022-02-02
Fix explicit repo solver

NathanReb:
It used to prefer older versions over newer ones. This was due to the solver context implementation!

### DIFF
--- a/lib/opam_solve.ml
+++ b/lib/opam_solve.ml
@@ -327,6 +327,7 @@ module Multi_dir_context :
                 Opam.Pp.package_name name Opam.Pp.version version);
           acc)
 
+  (** Candidates must be returned in decreasing preference order *)
   let candidates t name =
     let map =
       List.fold_left t ~init:OpamPackage.Version.Map.empty
@@ -334,7 +335,10 @@ module Multi_dir_context :
           let candidates = candidates dir_context name in
           merge_candidates ~name acc candidates)
     in
-    OpamPackage.Version.Map.bindings map
+    (* The bindings of the version map will be sorted by increasing version.
+       We reverse it so it's in decreasing version order for the solver to pick
+       the highest satisfying version. *)
+    List.rev (OpamPackage.Version.Map.bindings map)
 
   let user_restrictions t pkg = user_restrictions (List.hd t) pkg
 
--- a/test/bin/local-solver-picks-higher-version.t/repo/packages/base-bigarray/base-bigarray.base/opam
+++ b/test/bin/local-solver-picks-higher-version.t/repo/packages/base-bigarray/base-bigarray.base/opam
@@ -1,0 +1,6 @@
+opam-version: "2.0"
+maintainer: "https://github.com/ocaml/opam-repository/issues"
+description: """
+Bigarray library distributed with the OCaml compiler
+"""
+
--- a/test/bin/local-solver-picks-higher-version.t/run.t
+++ b/test/bin/local-solver-picks-higher-version.t/run.t
@@ -52,6 +52,5 @@ opam-monorepo solver should pick a.0.2 here:
   ==> Querying opam database for their metadata and Dune compatibility.
   ==> Calculating exact pins for each of them.
   ==> Wrote lockfile with 1 entries to $TESTCASE_ROOT/test.opam.locked. You can now run opam monorepo pull to fetch their sources.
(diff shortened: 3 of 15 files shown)


NathanReb:
I did this TDD style: added the failing test first, then the fix.

Also to keep the history clean and buildable, the test does not actually fail in the first commit but the test explanatory text makes things clear. The fix also updates the test output.

lib/opam_solve.ml
```diff
@@ -327,14 +327,18 @@ module Multi_dir_context :
...
       List.fold_left t ~init:OpamPackage.Version.Map.empty
         ~f:(fun acc dir_context ->
           let candidates = candidates dir_context name in
           merge_candidates ~name acc candidates)
     in
-    OpamPackage.Version.Map.bindings map
+    (* The bindings of the version map will be sorted by increasing version.
+       We reverse it so it's in decreasing version order for the solver to pick
+       the highest satisfying version. *)
+    List.rev (OpamPackage.Version.Map.bindings map)
```
Leonidas-from-XIV, line 341: I looked up because I was confused by `bindings` to have an order (since I would expect either a random order or an order that is somehow dependent on insertion order), but indeed that's what OCaml's map does by default:

> Return the list of all bindings of the given map. The returned list is sorted in increasing order of keys with respect to the ordering `Ord.compare`, where `Ord` is the argument given to `Map.Make`.

I guess this is an artifact on how it is implemented but given this is documented I guess it is part of the implementation guarantees so it should be fine!

test/bin/local-solver-picks-higher-version.t/repo/packages/base-bigarray/base-bigarray.base/opam
```diff
@@ -0,0 +1,6 @@
+opam-version: "2.0"
```
Leonidas-from-XIV, line 1: Maybe these could be somehow symlinked? I expect these base packages to be part of a lot of tests so I am worried about them filling up the repo with copies (that might diverge as well).

test/bin/local-solver-picks-higher-version.t/run.t
```diff
@@ -0,0 +1,57 @@
...
+
+opam-monorepo solver should pick a.0.2 here:
+
+  $ opam-monorepo lock
+  ==> Using 1 locally scanned package as the target.
+  ==> Found 9 opam dependencies for the target package.
+  ==> Querying opam database for their metadata and Dune compatibility.
+  ==> Calculating exact pins for each of them.
+  ==> Wrote lockfile with 1 entries to $TESTCASE_ROOT/test.opam.locked. You can now run opam monorepo pull to fetch their sources.
+  $ cat test.opam.locked | grep "\"a\""
```
Leonidas-from-XIV, line 55: Maybe grep for `"a" {` to make "sure" that this is the `"a"` from the dependencies (or well, something like `"a"\s+\{`)

Leonidas-from-XIV (commented):
I quite like the TDD style to show that there is an error, show a test and then fix it. Though it is only fun when the error is easy to reproduce :-)

test/bin/local-solver-picks-higher-version.t/repo/packages/base-bigarray/base-bigarray.base/opam
NathanReb: Yeah it would be great, would you be okay if I was to do it in a separate PR?

I'll have to experiment a bit with symlinks and dune rules. The latter would likely be better and more portable of course!

test/bin/local-solver-picks-higher-version.t/run.t
NathanReb: I agree it would be nice to get rid of the other one from the `x-opam-monorepo-duniverse-dirs`. The only thing I was afraid of was that if we end up with `"a"` in the deps, it would show nothing, which is more confusing but now that I'm thinking about it, it's impossible to have that in the lockfile so I'll switch to your suggestion!

test/bin/local-solver-picks-higher-version.t/repo/packages/base-bigarray/base-bigarray.base/opam
Leonidas-from-XIV: Yeah, sure, we can also do that.

Leonidas-from-XIV (approved):

PR #254, opened 2022-02-03
Generate base packages rather than copying them in each test

NathanReb:
This removes `minimal-repo` and all of its copies and add an executable, called from the cram test directly, that generates it instead.


Leonidas-from-XIV (approved):
I was wondering whether there isn't a better solution of not inlining the data into the executable and copying the files instead. But I guess that could be difficult because one would still require to set up a dependency on the folder containing the repos to make dune copy the opam repo into the `_build` folder.

So I think this should be fine.

PR #266, opened 2022-03-01
Update lockfile

NathanReb:
This updates our lockfile to use the latest version of our dependencies.

I added an upper bound to dune to prevent us from upgrading to dune 3 until we have confidence that our users and most of the ecosystem is compatible.

While we don't require it, this lockfile and our codebase are compatible with dune 3 and can be built with it.

### DIFF
--- a/opam-monorepo.opam.locked
+++ b/opam-monorepo.opam.locked
@@ -6,27 +6,27 @@ depends: [
   "alcotest" {= "1.5.0"}
   "angstrom" {= "0.15.0"}
   "astring" {= "0.8.5+dune"}
-  "base" {= "v0.14.1"}
+  "base" {= "v0.15.0"}
   "base-bigarray" {= "base"}
   "base-bytes" {= "base"}
   "base-threads" {= "base"}
   "base-unix" {= "base"}
-  "bigarray-compat" {= "1.0.0"}
+  "bigarray-compat" {= "1.1.0"}
   "bigstringaf" {= "0.8.0"}
-  "bos" {= "0.2.0+dune"}
+  "bos" {= "0.2.1+dune"}
   "cmdliner" {= "1.0.4+dune"}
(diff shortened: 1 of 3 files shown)


opam-monorepo.opam.locked
```diff
@@ -93,19 +93,19 @@ pin-depends: [
...
   ]
   [
     "bigstringaf.0.8.0"
     "https://github.com/inhabitedtype/bigstringaf/archive/0.8.0.tar.gz"
   ]
   [
-    "bos.0.2.0+dune"
-    "https://github.com/dune-universe/bos/archive/v0.2.0+dune.tar.gz"
+    "bos.0.2.1+dune"
+    "git+https://github.com/dune-universe/bos.git#e8fa18765d4538bc5d6cc1501891d0cde3d1a797"
```
Leonidas-from-XIV, line 108: Shouldn't we better have a tarball here?

Leonidas-from-XIV (approved):
It is most likely fine, my only concern is having `bos` as a git dependency.

opam-monorepo.opam.locked
NathanReb: good point, I'll go and update the port!

Leonidas-from-XIV: I'm a bit confused, because I just checked the port and we have a regular tarball of 0.2.1+dune in our repo, which is also referenced by opam-overlays correctly. Maybe you have a pin that is interfering?

NathanReb:
Just rebased this, I guess we updated the bos port upstream to use a tarball as I don't have any pin setup and locking again fixed this!

NathanReb:
Note that I temporarily added an upper bound to cmdliner to do the upgrade separately!

Leonidas-from-XIV (approved):

PR #268, opened 2022-03-08
Check OCaml version of lockfile against the switch version

### LINKED ISSUE #267: Check the compiler version when pulling
When `opam-monorepo pull` is called it just pulls the tarballs from the lock file. What it does not do is to check the OCaml version, which can lead to issues where the locked packages require a newer compiler than is available in the switch, thus failing to compile with a syntax error of whatever code failed to compile on the current switch.

Since the ocaml-base-compiler version is noted in the lockfile, we could show an error/warning if the version of the compiler in the current switch does not match what is in the lockfile to improve errors.

Leonidas-from-XIV:
Closes #267.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,8 @@
   `opam-monorepo`. To control this a new optional Opam file field,
   `x-opam-monorepo-opam-provided` is introduced. Its value is a list of package
   names that are to be excluded from being pulled (#234, @Leonidas-from-XIV)
+- Show an error message when the OCaml version of the lock file does not match
+  the OCaml version of the switch (#267, #268, @Leonidas-from-XIV)
 
 ### Changed
 
--- a/cli/pull.ml
+++ b/cli/pull.ml
@@ -65,9 +65,30 @@ let run (`Yes yes) (`Root root) (`Lockfile explicit_lockfile)
         Common.filter_duniverse ~to_consider:duniverse_repos duniverse
       in
       let* () = check_dune_lang_version ~yes ~root in
-      OpamGlobalState.with_ `Lock_none (fun global_state ->
-          Pull.duniverse ~global_state ~root ~full
-            ~trim_clone:(not keep_git_dir) duniverse)
+      OpamGlobalState.with_ `Lock_none @@ fun global_state ->
+      let* locked_ocaml_version =
+        Lockfile.ocaml_version lockfile
+        |> Result.of_option ~error:(`Msg "OCaml compiler not in lockfile")
+      in
+      OpamSwitchState.with_ `Lock_none global_state @@ fun switch_state ->
+      let switch_ocaml_version =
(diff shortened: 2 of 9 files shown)


CHANGES.md
```diff
@@ -8,6 +8,8 @@
 - Show an error message when the solver can't find any version that satisfies
   the requested version constraint in the user's OPAM file (#215, #248,
   @Leonidas-from-XIV)
+- Show an error message when the OCaml version of the lock file does not match
+  the OCaml version of the switch (#265, #<PR_NUMBER>, @Leonidas-from-XIV)
```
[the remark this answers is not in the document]
Leonidas-from-XIV:
```suggestion
  the OCaml version of the switch (#267, #268, @Leonidas-from-XIV)
```

cli/pull.ml
```diff
@@ -65,9 +70,25 @@ let run (`Yes yes) (`Root root) (`Lockfile explicit_lockfile)
...
+      in
+      OpamSwitchState.with_ `Lock_none global_state @@ fun switch_state ->
+      let switch_ocaml_version =
+        OpamSwitchState.get_package switch_state Config.compiler_package_name
+        |> OpamPackage.version
+      in
+      (match version_is_at_least locked_ocaml_version switch_ocaml_version with
+      | true -> ()
+      | false ->
+          Logs.warn (fun l ->
```
NathanReb, line 86: How about printing the warning after the pull? I think it will make it clearer that the pull was successful and that this requires folloe up actions by the user rather than it being an opam-monorepo failure.

cli/pull.ml
```diff
@@ -65,9 +70,25 @@ let run (`Yes yes) (`Root root) (`Lockfile explicit_lockfile)
...
+              l "OCaml versions do not match: %a in lockfile, but %a in switch"
```
NathanReb, line 87: It might be useful to suggest how to fix this along with the diagnostic, how about something along those lines:
```
You might want to upgrade the compiler in your switch:
opam install ocaml.%a --update-invariant
```

NathanReb (commented):
It looks great! This is a very nice addition and makes for a much better user experience. Thanks for taking care of this!

I wonder if we shouldn't also check the dune version here but I guess it's not as necessary as the error you'd get would likely tell you that a higher dune version is required. In the end I believe this will be better addressed by having a command to setup the dev environment entirely!

cli/pull.ml
Leonidas-from-XIV: I think this is a good idea indeed! Also makes the warning more visible.

cli/pull.ml
Leonidas-from-XIV: I didn't know this was possible on a switch with a locked invariant — that's a great idea, I added it!

NathanReb (approved):

PR #269, opened 2022-03-17
Prevent unecessary selection of beta versions

### LINKED ISSUE #265: Beta version are selected by the solver again
A few months ago we added some logic to prevent the solver from selecting beta versions of the compiler but due to recent changes in the package and release model for OCaml betas, it's not working anymore and running `opam monorepo lock` without upper bound currently selects `4.14.0` beta versions.

NathanReb:
Fixes #265 !

Starting recently, beta versions of the compiler are released as `ocaml-base-compiler` and not `ocaml-variant` anymore. This had the side effect of making our solver pick beta versions if the constraints allowed it.

This fixes this issue but making the solver prefer regular versions over ones flagged with `avoid-version` in their opam metadata.

Such versions can still be selected when no regular version match the requirements.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,6 +30,8 @@
 
 - Fix a bug where a package which had a single version that built with dune and got selected by the solver
   would be reported has having no version building with dune. (#245, @Leonidas-from-XIV)
+- Fix the solver so it does not select beta versions of the compiler unless it
+  has to. (#269, @NathanReb)
 
 ### Removed
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,8 +30,8 @@
 
 - Fix a bug where a package which had a single version that built with dune and got selected by the solver
   would be reported has having no version building with dune. (#245, @Leonidas-from-XIV)
-- Fix the solver so it does not select beta versions of the compiler unless it
-  has to. (#269, @NathanReb)
+- Fix the solver so it does not select beta versions of the compiler unless
+  forced to by version constraints or `--ocaml-version`. (#269, @NathanReb)
 
 ### Removed
 
--- a/lib/opam_solve.ml
+++ b/lib/opam_solve.ml
@@ -152,6 +152,15 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
                    in
                    (version, Ok opam_file))
 
+  let sort_candidates versions =
+    let avoid_versions, regular_versions =
+      List.partition versions ~f:(fun (_version, opam_res) ->
+          match opam_res with
+          | Ok opam -> Opam.avoid_version opam
+          | Error _ -> false)
+    in
+    regular_versions @ avoid_versions
+
   let candidates
       {
         base_context;
@@ -170,6 +179,7 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
         Base_context.candidates base_context name
         |> filter_candidates ~allow_jbuilder ~require_dune ~name
         |> remove_opam_provided ~opam_provided
+        |> sort_candidates
 
   let user_restrictions { base_context; _ } name =
     Base_context.user_restrictions base_context name
--- a/lib/opam_solve.ml
+++ b/lib/opam_solve.ml
@@ -152,7 +152,7 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
                    in
                    (version, Ok opam_file))
 
-  let sort_candidates versions =
+  let demote_candidates_to_avoid versions =
     let avoid_versions, regular_versions =
       List.partition versions ~f:(fun (_version, opam_res) ->
           match opam_res with
@@ -179,7 +179,7 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
         Base_context.candidates base_context name
         |> filter_candidates ~allow_jbuilder ~require_dune ~name
         |> remove_opam_provided ~opam_provided
-        |> sort_candidates
+        |> demote_candidates_to_avoid
 
   let user_restrictions { base_context; _ } name =
     Base_context.user_restrictions base_context name
--- a/test/bin/compiler-beta.t/run.t
+++ b/test/bin/compiler-beta.t/run.t
@@ -103,8 +103,8 @@ should pick 4.13.1:
 
   $ opam-monorepo lock test > /dev/null
   $ opam show --no-lint -fdepends ./test.opam.locked | grep ocaml
-  "ocaml" {= "4.14.0"}
-  "ocaml-base-compiler" {= "4.14.0~beta1"}
+  "ocaml" {= "4.13.1"}
+  "ocaml-base-compiler" {= "4.13.1"}
   "ocaml-config" {= "2"}
   "ocaml-options-vanilla" {= "1"}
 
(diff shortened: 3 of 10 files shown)


lib/opam_solve.ml
```diff
@@ -90,12 +90,22 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
             else (version, Error Non_dune))
       versions
 
+  let sort_candidates versions =
```
Leonidas-from-XIV, line 93: I am not sure "sort" is the right term for it, given it doesn't really sort them but rather reorders them.

test/bin/compiler-beta.t/run.t
```diff
@@ -0,0 +1,132 @@
+We have a simple project that only depends on OCaml
```
Leonidas-from-XIV, line 1: Since the code doesn't do anything compiler-specific (which is a good thing!) maybe it would make sense to just set up dummy packages? Dummy packages can have more monimal `opam` files that are easier to grasp and review.

test/bin/compiler-beta.t/run.t
```diff
@@ -0,0 +1,132 @@
...
+
+  $ cat test.opam
```
Leonidas-from-XIV, line 3: I think printing the opam files entirely is not great, because the files are in Git anyway so the user can take a look at them. Rather I would suggest using the opam tool to print the interesting bit of the file, so the file can be modified if something unrelated needs to change without disrupting all the tests.

test/bin/compiler-beta.t/run.t
```diff
@@ -0,0 +1,132 @@
...
+  ]
+
+We use the minimal repository which includes OCaml 4.13.1
+
+  $ gen-minimal-repo
+
+We also use an extra repository which happens to contain the packages of the
+ongoing beta release of OCaml 4.14.1
+
+  $ cat repo-with-beta-ocaml/packages/ocaml/ocaml.4.14.0/opam
```
Leonidas-from-XIV, line 21: Similar as above. This adds 80 lines of which possibly only like 71 is important.

test/bin/compiler-beta.t/run.t
```diff
@@ -0,0 +1,132 @@
...
+     (enabled by default).
+     Please file a bug report at https://github.com/ocaml/opam-repository/issues"
+    {failure & jobs > 1}
+    "You can try installing again including --jobs=1
+     to force a sequential build instead."
+    {failure & jobs > 1 & opam-version >= "2.0.5"}
+  ]
+
+Our package does not define an upper bound on ocaml but the beta should not be
+selected by default if there exist another solution, therefore here the solver
```
Leonidas-from-XIV, line 101:
```suggestion
selected by default if there exists another solution, therefore here the solver
```

test/bin/compiler-beta.t/run.t
```diff
@@ -0,0 +1,132 @@
...
+should pick 4.13.1:
```
Leonidas-from-XIV, line 102: This is fragile, it will break if a 4.13.2 is released.

Leonidas-from-XIV (commented):
I think this is a nice solution and I especially like how flexible this is to not only work for the compiler but also for any other package (like alpha/beta/release candidate releases etc) that has that flag set. My comments are mostly about making the test better/more reliable.

test/bin/compiler-beta.t/run.t
NathanReb: That would work but I like the fact that it shows it solves a real world problem, with real world metadata.

lib/opam_solve.ml
Leonidas-from-XIV: Maybe `demote_candidates_to_avoid`? I would say it doesn't really sort them (because it requires the candidates to be sorted, otherwise we could conclude that the identity function is a sorting function that just requires the the input to be sorted), it more splits or partitions them.

test/bin/compiler-beta.t/run.t
Leonidas-from-XIV: This file is short but it also includes a bunch of boilerplate, like the way the repos are set up which has nothing to do with the test and were it to change the test would need to be adapted for reasons that have nothing to do with the test itself.

test/bin/compiler-beta.t/run.t
Leonidas-from-XIV: I agree in principle, but in practice I find these files very long and distracting to read (if there is something wrong, do you want to go through the whole file and try to understand what each of these stanzas does and how it interacts with the others?), so to review it I would need to read through every stanza, understand it, see if it is important etc. I would prefer the tests to concentrate on one aspect and show it working.

Kinda like we want to use minimal examples for tests, so reading and modifying these is easy instead of dealing with accidental complexity.

test/bin/compiler-beta.t/run.t
NathanReb: You're right that printing those files entirely does not bring much value here. I'll remove that in favor of something shorter while still documenting those were taken from upstream opam repo.

test/bin/compiler-beta.t/run.t
NathanReb: It won't because it uses our minimal repo, not upstream!

test/bin/compiler-beta.t/run.t
NathanReb: I'll remove them but I don't see any reason why they would need to change since they are in fact very specific to this test and it's actually quite important to know what metadata will be used by the solver.

That's one of the reason why you can tell that it will always pick `4.13.1` below for instance.

test/bin/compiler-beta.t/run.t
NathanReb: I think here it's important that it also checks that it works with the `ocaml` and `ocaml-base-compiler` split. If we change this test we can still break this behaviour while having the dummy tests pass. I'm happy to extract another test just for the avoid-version if you'd like but this test will stay as a regression test.

lib/opam_solve.ml
NathanReb: Good name indeed, I'll change it!

NathanReb:
Rebased!

CHANGES.md
```diff
@@ -30,6 +30,8 @@
 
 - Fix a bug where a package which had a single version that built with dune and got selected by the solver
   would be reported has having no version building with dune. (#245, @Leonidas-from-XIV)
+- Fix the solver so it does not select beta versions of the compiler unless it
+  has to. (#269, @NathanReb)
```
Leonidas-from-XIV, line 34: Could you reword this a little? The current wording is a bit unclear, the reader is probably wondering "when does it have to then?". Specifying it like "it does not select beta version of the compiler unless forced by version constraints or on the command line" could be clearer.

Leonidas-from-XIV (approved):
Thanks for rebasing. I think this is an important fix, the code I'm :100: fine with. I'll leave it to you to reword the changelog entry to be a bit clearer, otherwise it is good to go :ship:

CHANGES.md
NathanReb: Yeah sure, it is not very clear indeed!

NathanReb: I think it's better know so I'll merge don't hesitate if you think it's still not clear enough, I'll be happy to update it separately!

We should do a brief review of the changelog before the release by the way to make sure everything is clear.

Leonidas-from-XIV: A review sounds good. Sometimes the entries make sense when writing the PR but afterwards it might be less so.

PR #271, opened 2022-03-24
RFC: Manage duniverse manual

Leonidas-from-XIV:
We discussed what the preferable way was but haven't written it down in an comprehensive way so that users understand the Pros and Cons of each solution.

This PR adds a new page to the manual to discuss what we think is the best way to run `opam-monorepo` and how to deal with the caveats.

Suggestions welcome!


doc/manage-duniverse.mld
```diff
@@ -0,0 +1,116 @@
...
+lock] will create a [.locked] opam file in addition to the original opam file.
+Calling [opam monorepo pull] will create a [duniverse] folder into which it
+will unpack the sources of the packages that are mentioned in the lockfile.
+
+{2 Not locked}
+
+While this is not a [opam-monorepo] workflow per se, projects that do not use
+[opam-monorepo] can still be used with [opam-monorepo]. As such every
+contributor is responsible for creating and maintianing their own lockfile and
+duniverse.
```
samoht, line 22: Do you mean calling `opam lock` or sourcing all your dependencies manually (with a script that generates https://github.com/janestreet/universe)?

doc/manage-duniverse.mld
```diff
@@ -0,0 +1,116 @@
...
+{- No opt-in to [opam-monorepo] required, every developer can chose to use it
+   or not}
+}
+
+Unfortunately, in this approach the disadvantages dominate:
+
+{ul
+{- Everyone might end up with different packages in their duniverse, thus low
+   reproducability}
+{- Everyone needs to have opam-monorepo installed}
```
samoht, line 36: Not sure to understand why do you need `opam-monorepo` installed here :-)

doc/manage-duniverse.mld
```diff
@@ -0,0 +1,116 @@
...
+{- The git repostitory does not grow by much, only one file}
+{- The packages to be installed are reproducible}
+{- Installation on the packages regularly into the Opam switch is still
+   possible}
+}
+
+Disadvantages might be
+
+{ul
+{- Everyone needs to have opam-monorepo installed to pull the dependencies}
```
samoht, line 60: I guess you can still use the lock file with opam? Also `opam-monorepo` being a plugin, it could automatically be installed by `opam` when you call it.

doc/manage-duniverse.mld
```diff
@@ -0,0 +1,116 @@
...
+{- Can be used offline}
+}
+
+Yet unfortunately it also has some disadvantages:
+
+{ul
+{- Contributors unaware of [opam-monorepo] might wonder where the sources
+   come from}
+{- Can lead to lots of files in repository}
+{- Potentially complicated merge conflicts if multiple contributors update
```
samoht, line 89: This is similar to multiple users editing the lock file simultaneously.

doc/manage-duniverse.mld
samoht: To complete that list :
1. users have to have opam installed *and* set up (with `opam init -ay`, a proper environment, etc)
2. the more dependencies you have, the more likely a network issue can occur and make `pull` fail.

doc/manage-duniverse.mld
```diff
@@ -0,0 +1,116 @@
...
+{- Users might modify the committed sources, yet those patches will be
+   overwritten when [opam monorepo pull] is run, leading to changes
+   disappearing}
+{- Imported packages can interact with version control in unexpected ways,
+   see {!label-git-add}}
+}
+
+For ease of collaboration and to keep the repositories manageable
+[opam-monorepo] maintainers recommend against this approach and suggest the
+approach in {!label-lockfile-in-git} instead.
```
samoht, line 100: I do not agree: there are valid use-cases for committing vendor code to the repositories, let's not try to dictate their requirements to our users. Would be much better to just reckon this, and point to possible issues and fixes (as the next paragraph does well).

doc/manage-duniverse.mld
```diff
@@ -0,0 +1,116 @@
...
+recommendations.
+
+{1 Methods}
+
+We assume the opam file is committed to the repository. Running [opam monorepo
+lock] will create a [.locked] opam file in addition to the original opam file.
+Calling [opam monorepo pull] will create a [duniverse] folder into which it
+will unpack the sources of the packages that are mentioned in the lockfile.
+
+{2 Not locked}
```
samoht, line 17: Maybe do this in the reverse order: start with the lock-file solution, then the vendoring version and finally the free-form one.

doc/manage-duniverse.mld
samoht: btw, the MirageOS tool is doing `opam-monorepo lock && opam-monorepo pull`, which is a perfectly valid use-case for the desired workflow (set-up a dune workspace to perform cross-compilation). Would also be nice to not dismiss this use-case if possible.

Leonidas-from-XIV: I meant the latter. You're right, I haven't thought of this usecase, I will add it.

doc/manage-duniverse.mld
```diff
@@ -0,0 +1,116 @@
...
+lockfile as well as the pulled sources.
+
+This appraoch has some advantages:
+
+{ul
+{- Only the person that creates the lockfile and duniverse must have
+   [opam-monorepo] installed}
+{- The repository can be immediately used after cloning as everything is
+   self-contained}
+{- Can be used offline}
```
samoht, line 80: another point here (which is both good and bad and you listed it only in the bad part :p): it's easy to quickly patch dependencies when you are trying to work collaboratively in a large-scale change. So it's a great way to experiment with new features like eio, or to test new releases, etc.

doc/manage-duniverse.mld
```diff
@@ -0,0 +1,116 @@
...
+}
+
+Yet unfortunately it also has some disadvantages:
+
+{ul
+{- Contributors unaware of [opam-monorepo] might wonder where the sources
+   come from}
```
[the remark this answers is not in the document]
Leonidas-from-XIV: Yes, but they need to be aware of `opam-monorepo` and the fact that the repository uses it. For a person that doesn't know about the tool, a populated duniverse folder is a lot of sources whose origin might be unclear and unclear how to update them. They might've been vendored manually/with a script, by unpacking tarballs in that folder.

I think it would be nice to somehow help users to figure out what the folder is and how to use it, a little bit like the comment in the dune-generated `opam` files that the file is generated and `dune-project` should be used instead. Maybe `opam-monorepo` should generate a `README.md` or something similar like this automatically which contains some instructions for contributors what the folder is and how to manage it?

doc/manage-duniverse.mld
Leonidas-from-XIV: A merge conflict of a single file (where the relevant parts are the versions) is in my experience much easier to resolve than a merge conflict of a source tree.

doc/manage-duniverse.mld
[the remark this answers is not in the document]
samoht: Generating a README.md in the `duniverse/` directory is indeed a great idea!

doc/manage-duniverse.mld
avsm: Indeed, irrespective of the recommendations of the opam-monorepo maintainers, I am going to continue checking in my code into duniverse/ :-)  The reason for this is that the benefit of this approach is somewhat understated and hidden away:
> The repository can be immediately used after cloning as everything is self-contained

The above is an absolutely massive benefit for newcomers to a project:
- it doesn't require opam to be installed, and, more crucially, doesn't require a working `opam init` to have happened.
- only `dune` is required for a build, and (since dune 2.0) the `dune` binary can be distributed independently of the dune libraries (which are vendored).   Thus a newcomer can `apt install ocaml-dune` and get on with things.
- it doesn't require opam-monorepo to remain working into the future.  I can `git checkout` an older revision of RWO and have a reasonable expectation that it'll build, for example.

This guide is an excellent start, but it is organised around "operational" uses of monorepo. But a reader of the document unfamiliar with opam-monorepo is going to start from a different mindset: they are going to wonder how best to serve various users of their projects and achieve outcomes such as "distribute this binary for people that don't know ocaml", to "make first contribution to the core tool", to "experienced user wants to patch a library and upstream the change".

If you organise things by "how deeply someone wants to get into the project", then the ordering of how you introduce vendoring, lockfiles and on-the-fly locking changes.

doc/manage-duniverse.mld
Leonidas-from-XIV: Yeah, I think this should say "everyone that wants to use opam-monorepo needs to have it installed" but it is an odd point especially in the advantages I stated that it is opt in so I just removed it to be less confusing.

doc/manage-duniverse.mld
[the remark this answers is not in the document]
Leonidas-from-XIV: Created an issue for this, because I believe this idea has potential: #272.

doc/manage-duniverse.mld
```diff
@@ -0,0 +1,213 @@
...
+
+Advantages include:
+
+{ul
+{- No opt-in to [opam-monorepo] required, every developer can chose to use it
+   or not}
+{- Easier management of dependencies over dependencies installed in opam
+   switches}
+{- No merge conflicts when multiple collaborators change versions}
+{- Having the dependencies built by Dune makes cross-compilation easier}
```
samoht, line 38: it also allows using lots of nice features of using `dune` at scale like the dune cache, or dune watch mode, etc.

doc/manage-duniverse.mld
```diff
@@ -0,0 +1,213 @@
...
+   reproducability}
+{- Not self-contained}
+}
+
+{3 Adoption}
+
+Using this workflow is very simple
+
+{[
+$ opam monorepo lock
```
samoht, line 54: unfortunately with the current setting this command is most likely to fail as you first need to add the opam-opervlays remote to the current switch (would be nice to simplify this at one point).

doc/manage-duniverse.mld
```diff
@@ -0,0 +1,213 @@
...
+   switches}
+{- Having the dependencies built by Dune makes cross-compilation easier}
+}
+
+Disadvantages might be
+
+{ul
+{- Everyone who wants to build the dependencies with Dune needs to have
+   [opam-monorepo] installed to pull the dependencies first}
+{- Not self-contained}
```
samoht, line 92: Maybe say something about the CI which should now support testing both the locked and non-locked versions?

doc/manage-duniverse.mld
```diff
@@ -0,0 +1,213 @@
...
+download all the dependencies via [opam monorepo pull] and then commit both the
+lockfile as well as the pulled sources.
+
+This appraoch has some advantages:
+
+{ul
+{- Only the person that creates the lockfile and duniverse must have
+   [opam-monorepo] installed}
+{- The repository can be immediately used after cloning as everything is
+   self-contained}
```
samoht, line 142: (maybe specify that you just need `dune` and `ocaml` installed - `opam`doesnt need to be installed nor configured)

doc/manage-duniverse.mld
Leonidas-from-XIV: Done. Originally I had it ordered "in order of commitment" for the repo but this might indeed be a better order.

doc/manage-duniverse.mld
NathanReb: As far as I know the CI has a single opam-monorepo workflow which is: if there is a monorepo lockfile, `pull` and then build.

It shouldn't be too hard to eventually detect that the duniverse is present and skip the first bit but it's not doing it at the moment.

doc/manage-duniverse.mld
```diff
@@ -0,0 +1,212 @@
...
+
+Another approach is very similar to the regular opam workflow, but in addition
+to having the [.opam] file committed to the repository, also the lockfile that
+[opam monorepo lock] generates is added.
+
+Advantages include:
+
+{ul
+{- The project repository only grows by one file, thus easy to adopt}
+{- The packages to be installed are reproducible between different machines
```
NathanReb, line 31: This items and the next one are advantages tied to the use of opam-monorepo itself and are common to the two described workflows.

We can probably remove them from here, eventually moving them elsewhere in the manual.

doc/manage-duniverse.mld
NathanReb: Isn't this covered by:
```
{- Easy to modify dependencies if required}
```

It might have been added after your initial comment, I'm mostly trying to figure out whether this comment is addressed or not.

doc/manage-duniverse.mld
NathanReb: To clarify: we should mention the CI indeed but also make sure users know exactly what it supports.

NathanReb: Also, the two items in this list derive from the same thing: dependencies are not packaged in the repository. I think they should be merged into one!

NathanReb (commented):
It looks good! Overall I think it can be simplified a little a bit, some sections are very similar between the two main workflows and can likely be merged.

doc/manage-duniverse.mld
Leonidas-from-XIV: I think I added this note later on after seeing @samoht's comment, just didn't mark as resolved because I was pondering pros and cons of doing so.

I think this workflow with "ejecting" changes is certainly interesting, but I am not sure using opam-monorepo (at least in its current form) is the optimal approach because it would require a complicated system track the changes. Especially given what opam-monorepo usually unpacks are tarballs, one would need to extract the changes, correlate them back to the git repo (which might not correspond to the tarball and also might have seen development since that particular release).

To me personally it seems more manageable to symlink or pin-depend or git-submodule a repo for iterating on changes to the dependencies than to modify files that opam-monorepo has vendored. That way it is much easier to differentiate between upstream and my changes and much easier to submit those upstream or have them in my own fork.

doc/manage-duniverse.mld
Leonidas-from-XIV: I've added a note. With the "explicit repo" feature we should be able to simplify this a bit by adding the opam-overlays repository on the opam-monorepo side automatically, without messing with the users switch. There's a discussion in #273.

PR #274, opened 2022-03-28
Write a README.md when pulling a project

### LINKED ISSUE #272: Generate documentation in the `duniverse/` folder
While discussing the possible ways to use opam-monorepo I raised the point that for someone who is not aware of `opam-monorepo` the `duniverse/` folder with all the sources can be confusing: how was it created, are there forked versions, why is it there, how to update packages from there etc.

To help with the confusion, `opam-monorepo` could generate a documentation file, e.g. a `README.md` in the duniverse folder automatically to document what it is and how to use it.

This has a number of advantages
 * When committing the `duniverse/` folder it serves as documentation, so even if the project has not documented its usage of `opam-monorepo` there will be some generic information in place. It could potentially also mention the versions of the packages and links to them.
 * Github would display it inline when opening the folder in the browser.
 * When not committing it it doesn't hurt, as it is just a small-ish text file.

Leonidas-from-XIV:
The wording can surely be amended, but I thought it would be convenient to generate such a documentation so users that see projects that have a `duniverse/` folder know what the deal with it is.

Closes #272


lib/pull.ml
```diff
@@ -50,6 +50,48 @@ let pre_pull_clean_up ~full ~duniverse_dir duniverse =
         Bos.OS.Dir.delete ~must_exist:false ~recurse:true
           Fpath.(duniverse_dir / dir))
 
+let duniverse_documentation =
+  {|\
+duniverse
+=========
```
NathanReb, line 56: Propbably better to use the `# duniverse` syntax here!

NathanReb (commented):
Looks great! This is really useful thanks for the idea and the work!

lib/pull.ml
Leonidas-from-XIV: Good point, I've changed this while rebasing :+1:

lib/pull.ml
```diff
@@ -50,6 +50,47 @@ let pre_pull_clean_up ~full ~duniverse_dir duniverse =
         Bos.OS.Dir.delete ~must_exist:false ~recurse:true
           Fpath.(duniverse_dir / dir))
 
+let duniverse_documentation =
+  {|\
```
patricoferris, line 54: Great idea adding this README :))

I think this little `\` might be a typo? I was using the upstream version of `opam-monorepo` (https://github.com/patricoferris/dream-multicore-repro/blob/main/duniverse/README.md) and noticed it.

NathanReb: Indeed, this syntax does not work with `{|` quoted strings! Good catch, let's open an issue!

PR #275, opened 2022-03-31
Add a basic test for the depexts

Leonidas-from-XIV:
We didn't have tests for the depexts, so here's a very basic one that shows that we capture transitive depexts correctly.

Writing a more complicated one is tricky because the packages installed depend on the OS and the state of the OS (whether they are installed or not) and I'd rather not maintain the package names that we use as dummies in the test because this will break on unknown platforms.

### DIFF
--- a/test/bin/depext.t/run.t
+++ b/test/bin/depext.t/run.t
@@ -1,0 +1,28 @@
+We want to make sure picking depexts works, even if we vendor packages
+
+We setup the default base repository
+
+  $ gen-minimal-repo
+
+Here we define a package test that depends on a package `b`:
+
+  $ opam show --no-lint --raw -fdepends ./a.opam
+  "dune" "b"
+
+Package `b` has a `depext` on a fantasy OS package. We deliberately pick a
+fantasy name here to make sure the behaviour is the same on all platforms.
+
+  $ opam show --no-lint --raw -fdepexts ./repo/packages/b/b.1/opam
+  "libfantasydependency"
+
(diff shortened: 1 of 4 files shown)


test/bin/depext.t/run.t
```diff
@@ -0,0 +1,28 @@
...
+We lock and expect the OS package to be part of the locked Opam file.
+
+  $ opam-monorepo lock a > /dev/null
+  $ opam show --no-lint --raw -fdepexts ./a.opam.locked
+  "libfantasydependency"
+
+Then we want to make sure `depext` works. Given the fantasy package does not
+exist, `depext` should not install any packages but still work successfully:
+
+  $ opam-monorepo depext --dry-run
```
NathanReb, line 27: I didn't even know we had a dry run there! Is it inherited from opam?

It would be nice if it output the list of packages to install or something similar.

NathanReb (approved):
It looks good, thanks!

There's already value in this test so feel free to merge but I think we should look into the depext thing still, it might be helpful for users too to get a bit of visibility before running it.

test/bin/depext.t/run.t
Leonidas-from-XIV: The problem is that it currently asks the OS what is installed so the output would be just the subset that needs to be installed - which is nice but pretty much unusable in a test that is supposed to be reproducible, and if the OS doesn't know the packages it just skips them. So if we were to change that and output this in `dry-run` it would do fundamentally different things than it would do in a regular run, which is not a great test to begin with.

PR #276, opened 2022-04-04
Update to cmdliner.1.1

NathanReb:
This PR updates the lockfile to use the latest version of cmdliner which introduced a new API and some deprecation warnings.

The upgrade turned some `1` exit codes in the test into `124` which stands for command line options parsing failure and was therefore misleading since those were clearly opam-monorepo errors.
I slightly changed how we turn results into exit codes to fix this but it also fixed other, pre-existing of occurences of false `124`.

### DIFF
--- a/bin/opam_monorepo.ml
+++ b/bin/opam_monorepo.ml
@@ -27,6 +27,43 @@ let init_opam () =
   OpamRepositoryConfig.init ();
   OpamStateConfig.init ~root_dir:root ()
 
+let default_run () = `Help (`Pager, None)
+
+let default = Cmdliner.Term.(ret (const default_run $ const ()))
+
+let info =
+  let open Cmdliner in
+  let doc = "the spice of build life" in
+  let sdocs = Manpage.s_common_options in
+  let man_xrefs = [ `Tool "dune"; `Tool "opam"; `Tool "git" ] in
+  let man =
+    [
+      `S Manpage.s_description;
+      `P
+        "The $(tname) plugin provides a convenient interface to bridge the \
+         $(b,opam) package manager with having a local copy of all the source \
+         code required to build a project using the $(b,dune) build tool.";
+      `P
+        "It works by analysing opam package metadata and calculating a set of \
+         URLs that can be downloaded or cloned into the local repository into \
+         a $(b,duniverse/) subdirectory. Once the external code has been \
+         pulled into the repository, a single $(b,dune build) command is \
+         sufficient to build the whole project in a standalone fashion, \
+         without opam being required. This is a particularly convenient way of \
+         publishing CLI tools to users who do not need the full power of opam.";
+      `P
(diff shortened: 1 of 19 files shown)


tmattio (approved):
LGTM!

I was thinking that we could start using let operators* to make the CLI code slightly more readable, but that's definitely not in scope for this PR 🙂 

```ocaml
module Syntax = struct
  let ( let+ ) t f = Term.(const f $ t)

  let ( and+ ) a b = Term.(const (fun x y -> x, y) $ a $ b)
end
```

cli/lock.ml
```diff
@@ -548,12 +548,12 @@ let info =
          instructions there to add dune ports for the packages you need.";
     ]
   in
-  Term.info "lock" ~doc ~exits ~man
+  Cmd.info "lock" ~doc ~exits ~man
 
 let term =
   let open Term in
```
Leonidas-from-XIV, line 554: Maybe while at it, avoid the `open Term` and use `Term.(const run …)`? With multiple modules named `Term` it becomes a bit less obvious what should be opened and what not. `pull.ml` does this that way and I think it is more readable and understandable.

dune-project
```diff
@@ -22,7 +22,7 @@ code required to build a project using the dune build tool.")
   dune-build-info
   base
   bos
-  (cmdliner (< 1.1.0))
+  cmdliner
```
Leonidas-from-XIV, line 25: I think we need to have a minimum constraint here:

```suggestion
  (cmdliner (>= 1.1.0))
```

bin/opam_monorepo.ml
```diff
@@ -27,6 +27,43 @@ let init_opam () =
...
+         a $(b,duniverse/) subdirectory. Once the external code has been \
+         pulled into the repository, a single $(b,dune build) command is \
+         sufficient to build the whole project in a standalone fashion, \
+         without opam being required. This is a particularly convenient way of \
+         publishing CLI tools to users who do not need the full power of opam.";
+      `P
+        "You can access the functionality directly via the $(i,monorepo-lock) \
+         and $(i,monorepo-pull) commands,";
+      `P
+        "Also see $(i,https://github.com/avsm/platform) for an example of a \
```
Leonidas-from-XIV, line 58: I know this was copied from the other file, but this repo hasn't been updated in a while, maybe we should point at a different example?

Leonidas-from-XIV (commented):
Very nice. The error code conversion in older versions of Cmdliner always felt a bit wonky so I think we're doing better now.

The only change that I think should be looked into is the cmdliner minimum constraint, the rest looks very good!

bin/opam_monorepo.ml
NathanReb: Yeah, let's open an issue about this!

Leonidas-from-XIV (approved):

PR #282, opened 2022-04-14
Improve vendor marking so that it requires no opam configuration

### LINKED ISSUE #257: Undefined boolean filter value: vendor
Hello! Thank you for `opam-monorepo`, I've used it in quite a few projects including most recently in [a multicore monorepo](https://github.com/patricoferris/ocaml-multicore-monorepo), where I pinned to the latest `main` branch of opam-monorepo.

When doing this I ran into the following problem. I'm trying to provide a cross-platform monorepo given that `dune` itself should be able to build the correct tools depending on the OS. At the moment this is not possible it would seem given that the packages opam-monorepo chooses are based on what can be installed on your OS. Luckily, for the multicore monorepo building on Linux also sets up an environment that could be built on macOS (or windows). In particular this is because they all support the luv backend for Eio. Building the monorepo on macOS means the io_uring backend doesn't get pulled in and switching to linux this causes issues (see this CI log: https://github.com/patricoferris/ocaml-multicore-monorepo/runs/5157933461?check_suite_focus=true). Maybe this is just a fundamental limitation or maybe I'm doing something wrong? 

I started developing on macOS so only found this problem when adding CI, I then tried to fix this by doing a `lock + pull` on a Linux machine. In order to do this I had to try and setup a switch that looked like the one on macOS which I believe should be achievable with `opam install . --locked` but with the changes in #237 I got:

```
$ opam --version
2.1.0
$ opam install . --locked
Fatal error: Undefined boolean filter value: vendor
```

Perhaps I've missed something to set this up correctly?

All in all, thanks for the great tool! I plan to use it a lot more for personal projects I want to work for a long time :))

NathanReb:
Fixes #257 

Opam has a way to allow undefined variables to be assigned a default value.
Our initial assumption was that it was doing this for all variables but it was wrong. Variables need to be applied the prefix, unary `?` operator.

This PR adds the operator before the `vendor` variable when writing the lockfile. It also expects the variable filter to be written this way when reading it back.

### DIFF
--- a/lib/lockfile.ml
+++ b/lib/lockfile.ml
@@ -115,11 +115,11 @@ module Depends = struct
           ( name,
             And
               ( Atom (Constraint (`Eq, FString version)),
-                Atom (Filter (FIdent ([], var, None))) ) )
+                Atom (Filter (FDefined (FIdent ([], var, None)))) ) )
       | Atom
           ( name,
             And
-              ( Atom (Filter (FIdent ([], var, None))),
+              ( Atom (Filter (FDefined (FIdent ([], var, None)))),
                 Atom (Constraint (`Eq, FString version)) ) )
         when variable_equal var Config.vendor_variable ->
           let version = OpamPackage.Version.of_string version in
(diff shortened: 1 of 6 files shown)


Leonidas-from-XIV (approved):
That's a pleasantly simple fix, looks good, let's get it merged!

tmattio (commented):
I'm not familiar enough with the codebase to provide a useful review on this one, so by any means, if this looks good to @Leonidas-from-XIV, don't wait for my stamp to merge this 🙂

PR #283, opened 2022-04-14
Relax constraint on dune to support dune 3

tmattio:
The project seems to build fine on my machine with dune 3. I had some test failures that seemed unrelated (incompatible grep CLI used in cram tests)


NathanReb:
It's entirely possible. I added this upper bound a little while ago when some dependencies would not build with dune 3.

NathanReb (approved):

PR #288, opened 2022-04-20
Use opam show --just-file in cram tests

NathanReb:
This speeds up the tests quite a bit and prevents it from trying to write anything which will help when sandboxed!

### DIFF
--- a/test/bin/compiler-beta.t/run.t
+++ b/test/bin/compiler-beta.t/run.t
@@ -1,6 +1,6 @@
 We have a simple project that only depends on OCaml
 
-  $ opam show --just-file --no-lint -fdepends ./test.opam
+  $ opam show --just-file -fdepends ./test.opam
   ocaml, dune
 
 We use the minimal repository which includes OCaml 4.13.1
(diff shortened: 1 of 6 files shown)


test/bin/compiler-beta.t/run.t
```diff
@@ -1,7 +1,7 @@
 We have a simple project that only depends on OCaml
 
-  $ opam show --no-lint -fdepends ./test.opam
-  "ocaml" "dune"
+  $ opam show --just-file --no-lint -fdepends ./test.opam
```
emillon, line 3:
```suggestion
  $ opam show --just-file -fdepends ./test.opam
```

I believe that `--no-lint` is now not necessary anymore

emillon (changes requested):

test/bin/compiler-beta.t/run.t
emillon: (in all these tests)

Leonidas-from-XIV (approved):
Ah, this is much better anyway.

emillon (approved):

PR #290, opened 2022-04-20
Relax ocaml compiler warning to allow ocaml-variants

Leonidas-from-XIV:
The warning introduced in #268 had the downside that it only worked when users would use `ocaml-base-compiler`. However, when using variants the compiler is not `ocaml-base-compiler` but `ocaml-variants`, as such the check needs to be relaxed to accept either of these options.

Sidenote: Our `List.mem` from `Stdext` does not support passing in a custom `equal` function, maybe a good idea to add?

Reported by @patricoferris

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -20,7 +20,7 @@
   `x-opam-monorepo-global-opam-vars` to make `lock` fully reproducible.
   (#250, #253, @NathanReb)
 - Show an error message when the solver can't find any version that satisfies
-  the requested version constraint in the user's OPAM file (#215, #248,
+  the requested version constraint in the user's OPAM file (#215, #248, #290
   @Leonidas-from-XIV)
 - Allow packages to be marked as being provided by Opam and not to be pulled by
   `opam-monorepo`. To control this a new optional Opam file field,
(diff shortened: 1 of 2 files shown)


NathanReb (commented):
I wonder if this shouldn't be made simpler and just compare `ocaml` versions. There are few things that can go wrong here:
- `ocaml-system` and `ocaml-base-compiler` should likely be treated as equal here otherwise one can get a warning when their switch is perfectly fit to build the project.
- `ocaml-variants` can have slightly different versions such as `4.14.0+trunk` which could be treated as equal in some circumstances
- It might make sense in certain situation to use an ocaml-variant even if the lockfile says `ocaml-base-compiler`, for instance if you want to try flambda.

With all this in mind I think just making sure `ocaml`'s version match would be enough.

Leonidas-from-XIV:
You might be right, it would be quite annoying to have to match OCaml versions across packages with all kinds of special cases. I've changed it so it only uses `ocaml`. There might be a point of moving the `ocaml_compiler_name` out of `config.ml` because it is only used for `pull` but in order no not sidetrack the 0.3 release with rework of this feature I think we can leave it as is.

I'll remove the patch and its revert before merging.

PR #293, opened 2022-04-26
minor line editing for ReadMe

christinerose:
Fixed just a few capitalisations and other formatting. 

Signed-off-by: Christine Rose <professor.rose@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ so that Dune will only build the artifacts you need from that folder.
 From that point you should be able to run `dune build` and `dune runtest` as you normally would and
 build your entire project from scratch!
 
-### `opam-monorepo lock`
+### The `lock` Command
 
 It's important to note that `opam monorepo lock` will only succeed if all of your non-virtual and
 non "base" dependencies (e.g., `ocaml` or `dune`) build with Dune (i.e., directly depend on the `dune`
--- a/README.md
+++ b/README.md
@@ -1,12 +1,12 @@
 [![Build Status](https://img.shields.io/endpoint?url=https%3A%2F%2Fci.ocamllabs.io%2Fbadge%2Focamllabs%2Fopam-monorepo%2Fmain&logo=ocaml)](https://ci.ocamllabs.io/github/ocamllabs/opam-monorepo) [![Documentation](https://img.shields.io/badge/doc-online-blue)](https://ocamllabs.io/opam-monorepo/opam-monorepo/index.html)
 
-# opam-monorepo
+# `opam-monorepo`
 
-`opam-monorepo` is an opam plugin designed to assemble standalone dune workspaces
+`opam-monorepo` is an opam plugin designed to assemble standalone Dune workspaces
 with your projects and all of their dependencies, letting you build it all from scratch
 using only `dune` and `ocaml`.
 
-[Documentation on opam-monorepo](https://ocamllabs.io/opam-monorepo/opam-monorepo/index.html) is available in the repository as well as available online.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,72 +16,77 @@ You can simply install it via opam in your current switch by running:
...
-[dune's documentation](https://dune.readthedocs.io/en/latest/dune-files.html#vendored-dirs-since-1-11))
-so that dune will only build the artifacts you need from that folder.
+[Dune's documentation](https://dune.readthedocs.io/en/latest/dune-files.html#vendored-dirs-since-1-11))
+so that Dune will only build the artifacts you need from that folder.
 
 From that point you should be able to run `dune build` and `dune runtest` as you normally would and
 build your entire project from scratch!
 
-### opam monorepo lock
+### `opam-monorepo lock`
```
NathanReb, line 56: This is not a typo and should indeed be:
```suggestion
### `opam monorepo lock`
```

README.md
```diff
@@ -16,72 +16,77 @@ You can simply install it via opam in your current switch by running:
...
+a reproducible way (i.e., independently of any change that might have happened on the upstream
+`opam-repository`) thanks to the `pin-depends`.
+You can use that property to your advantage by allowing one to choose between a "monorepo" or a
 regular opam workflow depending on the situation.
 
 You can also exclude packages from the set of packages to
-be vendored by `opam-monorepo`. To do so you can specify an additional field in
-your Opam file:
+be vendored by `opam-monorepo`. To do so, specify an additional field in
+your opam file:
```
NathanReb, line 89: Is there any particular reason why Dune and Git are capitalized but not opam?

README.md
NathanReb: Alternatively we could rename that section to something like this:
```suggestion
### The `lock` command
```

README.md
```diff
@@ -16,72 +16,77 @@ You can simply install it via opam in your current switch by running:
 opam install opam-monorepo
 ```
 
-Note that once it is installed you can invoke it as an opam command:
+Note that once it's installed you can invoke it as an opam command:
```
NathanReb, line 19: Out of curiosity, what's the rule about contraction in written English?

I personally use contractions when I write but I remember somebody telling me it shouldn't be used in written English. Seeing this, I'm now confused!

NathanReb (commented):
Thanks for the update! The README also needs a content refresh but that's already a good first step!

README.md
christinerose: Whichever is most accurate! Since `opam-monorepo` is generally in monospace, I wondered if it should be `opam-monorepo` lock if not `opam-monorepo lock` ... Although I like The `lock` Command, since heading, you can also capitalise Command. ... and Lock, if it's not incorrect. Even 'The Lock Command' ?

README.md
christinerose: Yes. Opam under the rules of English language _should be capitalised_, as it's a proper noun, but Thomas feels strongly about its official name being formatted in lower case. Dune and Git are also proper nouns, so should be capitalised when not referring to the command like `dune` and `git` . I've already spoken with the maintainers of Dune, and they agree with the Dune and `dune` formatting. As for Git, it's capitalised by its maintainers and across the web.

README.md
christinerose: I prefer them because it's more casual in most cases. The tone of our documentation, by directly talking to the reader in the second person (you, your) and us in the first person (we, our), I feel it's more natural to use contractions unless there needs to be particular emphasis.

Not using contractions gives it a more formal tone. 

Basically, it's a matter of preference. Both are correct.

NathanReb (approved):

PR #303, opened 2022-05-16
Fix OCaml lower bound

NathanReb:
We discovered during the previous release that some of our vendored dependencies are not compatible with 4.08 anymore.
This PR bumps the lower bound to 4.10 accordingly.

### DIFF
--- a/.ocamlformat
+++ b/.ocamlformat
@@ -1,3 +1,3 @@
 version=0.20.1
 profile=conventional
-ocaml-version=4.08
+ocaml-version=4.10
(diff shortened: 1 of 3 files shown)


Leonidas-from-XIV:
Can you update the `.ocamlformat` file as well? I doubt that there are any changes triggered by this but it is good to be consistent. Maybe `ocamlformat` could read the info automatically from somewhere (`dune-project`?) in the future.

NathanReb:
Done!

Leonidas-from-XIV (approved):

PR #304, opened 2022-05-16
Unpin opam-0install and update lock file

NathanReb:
There has been a release of opam-0install incorporating the changes we needed so we can unpin it now!


Leonidas-from-XIV (approved):
That's great news! Probably should to be rebased when #303 is merged.

PR #305, opened 2022-05-17
Minimal update of lock file

NathanReb:
This PR adds a `--minimal-update` flag to `lock` which allows for updating the lock file, aiming at minimum changes in the dependencies.

When the flag is passed, the target lockfile (i.e. the one that it will write the generated lock file to) must already exists. It parsed and the list of packages extracted from it. It builds a map `name -> version` that is then passed to the solver. The map is used by the context to promote those version to the top of the preference sorted list of candidates for the package.
This result in different versions only getting picked if no solution exist with the version form the previous lock file. New packages are picked as they normally would and package that are not required anymore are also dropped without issues.

In short the preference order allows us to do everything we wanted from that feature making this a very easy to implement yet extremely useful feature!

### DIFF
--- a/cli/lock.ml
+++ b/cli/lock.ml
@@ -269,8 +269,8 @@ let opam_provided_packages ~opam_monorepo_cwd local_packages target_packages =
     target_packages (Ok OpamPackage.Name.Set.empty)
 
 let calculate_opam ~source_config ~build_only ~allow_jbuilder
-    ~require_cross_compile ~local_opam_files ~ocaml_version ~target_packages
-    ~opam_provided =
+    ~require_cross_compile ~preferred_versions ~local_opam_files ~ocaml_version
+    ~target_packages ~opam_provided =
   let open Result.O in
   OpamGlobalState.with_ `Lock_none (fun global_state ->
       let* pin_depends = get_pin_depends ~global_state local_opam_files in
@@ -287,8 +287,8 @@ let calculate_opam ~source_config ~build_only ~allow_jbuilder
           let opam_env = extract_opam_env ~source_config global_state in
           let solver = Opam_solve.explicit_repos_solver in
           Opam_solve.calculate ~build_only ~allow_jbuilder
-            ~require_cross_compile ~local_opam_files ~target_packages
-            ~opam_provided ~pin_depends ?ocaml_version solver
+            ~require_cross_compile ?preferred_versions ~local_opam_files
+            ~target_packages ~opam_provided ~pin_depends ?ocaml_version solver
             (opam_env, local_repo_dirs)
           |> Result.map_error ~f:(interpret_solver_error ~repositories solver)
       | { repositories = None; _ } ->
@@ -298,8 +298,9 @@ let calculate_opam ~source_config ~build_only ~allow_jbuilder
                     (OpamSwitch.to_string switch_state.switch));
               let solver = Opam_solve.local_opam_config_solver in
               Opam_solve.calculate ~build_only ~allow_jbuilder
-                ~require_cross_compile ~local_opam_files ~target_packages
-                ~opam_provided ~pin_depends ?ocaml_version solver switch_state
+                ~require_cross_compile ?preferred_versions ~local_opam_files
+                ~target_packages ~opam_provided ~pin_depends ?ocaml_version
+                solver switch_state
               |> Result.map_error ~f:(fun err ->
                      let repositories = current_repos ~switch_state in
                      interpret_solver_error ~repositories solver err)))
@@ -413,6 +414,26 @@ let local_packages ~versions repo =
   log_local_packages local_packages_path;
   package_version_map ~versions local_packages_path
 
+let preferred_versions ~minimal_update ~root target_lockfile =
+  let open Result.O in
+  if not minimal_update then Ok None
+  else
+    let* lockfile_exists = Bos.OS.File.exists target_lockfile in
+    if not lockfile_exists then
+      Rresult.R.error_msgf "No lock file to upgrade, could not find %a" Fpath.pp
+        target_lockfile
+    else
+      let* lockfile =
+        Lockfile.load ~opam_monorepo_cwd:root ~file:target_lockfile
+      in
+      let depends = Lockfile.depends lockfile in
+      let name_to_version_map =
+        List.fold_left depends ~init:OpamPackage.Name.Map.empty
+          ~f:(fun acc { Lockfile.Depends.package; _ } ->
+            OpamPackage.Name.Map.add package.name package.version acc)
+      in
+      Ok (Some name_to_version_map)
+
 let extract_source_config ~opam_monorepo_cwd ~opam_files target_packages =
   let open Result.O in
   let target_opam_files =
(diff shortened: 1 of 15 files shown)


NathanReb:
Note that the documentation page I added is obviously incomplete but I think it's better to improve it iteratively than to wait to have a perfection version of it before merging it. I'm happy to add to it in a separate PR!

test/bin/minimal-update.t/run.t
```diff
@@ -0,0 +1,68 @@
+We have a simple project with a single package which depends on
+packages a and b:
```
Leonidas-from-XIV, line 2: Just a small thing: it would be nice if these were `a`, `b`, `c` in the readme, because sometimes it is hard to distinguish between the package `a` and the word "a".

lib/opam_solve.ml
```diff
@@ -175,6 +178,22 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
     in
     regular_versions @ avoid_versions
 
+  let prefer_version ~version versions =
```
Leonidas-from-XIV, line 181: `preferred_version` would be nicer here. Maybe even something like `let promote_version ~preferred_version candidate_versions =` as a analogue to `demote`.

lib/opam_solve.ml
```diff
@@ -175,6 +178,22 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
...
+    let rec aux acc l =
```
Leonidas-from-XIV, line 182: `aux` is a rather nondescript name and given the function is somewhat dense (it is not a loop) I would try to give it a better name. Also `l` could be replaced by `function`.

lib/opam_solve.ml
```diff
@@ -175,6 +178,22 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
...
+      match l with
+      | [] -> versions
+      | (v, opam) :: tl when OpamPackage.Version.equal v version ->
```
Leonidas-from-XIV, line 185:
```suggestion
      | ((version, _) as candidate) :: tl when OpamPackage.Version.equal version preferred_version ->
        candidate :: List.rev_append acc tl
```
feels more readable to me because it describes what the tuple is and comparing  `preferred_version` with `version` feels more understandable than comparing `version` with `v`.

lib/opam_solve.ml
```diff
@@ -52,6 +53,7 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
     require_dune : bool;
     opam_provided : OpamPackage.Name.Set.t;
     require_cross_compile : bool;
+    preferred_versions : OpamTypes.version OpamPackage.Name.Map.t option;
```
Leonidas-from-XIV, line 56: As mentioned in the other comments, maybe an empty map would be a cyclomatically less complex version.

I feel that there is no semantic difference in this case between no preferred versions in a map and no map - to me it feels like it expresses the same state ("no version preferred") but with the option there are now two ways to express this.

lib/opam_solve.ml
```diff
@@ -201,10 +221,15 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
           require_cross_compile
           && List.exists ~f:candidate_cross_compile candidates
         in
+        let preferred_version =
+          let open Option.O in
+          let* preferred_versions = preferred_versions in
```
Leonidas-from-XIV, line 226: I feel like this would be nicer simpler if it was just am empty map, saves one indirection and expresses the same semantically.

lib/opam_solve.ml
```diff
@@ -201,10 +221,15 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
...
+          OpamPackage.Name.Map.find_opt name preferred_versions
+        in
         filter_candidates ~allow_jbuilder ~must_cross_compile ~require_dune
           ~name candidates
         |> remove_opam_provided ~opam_provided
-        |> demote_candidates_to_avoid
+        |> sort_candidates_by_preference ~preferred_version
```
Leonidas-from-XIV, line 232: Is there a way to keep this as two steps, somehow? I feel like the double processing overhead is outweighted by the readability of having functions do one thing. Then `sort_candidates_by_preference` becomes basically just a quick pattern match.

cli/lock.ml
```diff
@@ -532,6 +557,18 @@ let require_cross_compile =
     (fun x -> `Require_cross_compile x)
     Arg.(value & flag & info ~doc [ "require-cross-compile" ])
 
+let minimal_update =
+  let doc =
+    "Generate a lock file with minimum dependency changes compared to the \
+     previous lock file. It can remove or add packages based on the dependency \
+     specification in the local opam files but will not upgrade or downgrade \
+     previously locked packages unless strictly necessary.\n\
+     Require a lock file to exist in the target lock file path."
```
Leonidas-from-XIV, line 566: I feel like specifying the mechanism how it does so e.g. "prefers to keep the versions as in the existing lock file unless strictly necessary" would make it feel a bit less magical - then it is clear how it tries to do minimal dependency changes without complicating the description too much.

What do you think?

Leonidas-from-XIV (commented):
I am also pleasantly surprised how little changes (at least from the core logic part) it is. Hence I mostly looked at that to suggest some improvements. Glad to see that the implementation matches how my hunch it could be implemented, which is great since that means that the complexity increase of the code is very manageable :+1:  

The flags & parsing and tests look good. It's a bunch of code but essentially not-exciting code and the compiler does a good job of making sure this checks out.

lib/opam_solve.ml
NathanReb: I personnaly quite dislike `function`, I prefer to declare all of my arguments so if you don't mind I'll stick to this.

The `aux` function here is just a recursive version of the main function with an extra `accumulator` arg. I find it to be a quite common pattern but I can rename it `recursive_prefer_version` if you'd like.

lib/opam_solve.ml
NathanReb: `preferred_version` would be confusing here as it just re-orders the list and that name would somehow indicate it will return a version.

I initially thought about naming it `promote_version` so let's go for this one which is indeed better than the current name!

lib/opam_solve.ml
Leonidas-from-XIV: It is more like `move_version_to_front`, because that describes what it is doing.

It is fine to dislike `function`, I can see the point because sometimes naming the arguments really helps, but in such case it would be good if the argument was called something else than `l`.

lib/opam_solve.ml
Leonidas-from-XIV: Oh, yeah, I meand `preferred_version` as the name of the named argument, not the name of the function. I think "promote" captures what the function is attempting to do pretty well!

NathanReb: I feel like if we use `promote_version` than naming the argument `~preferred_version` is a bit redundant.

cli/lock.ml
NathanReb: Yeah that's shorter and clearer! Let's change it!

lib/opam_solve.ml
NathanReb: Yeah, I went for an option when I started this as I thought it would allow more optimisation to avoid running useless operations over and over when there's no preferred_versions but it was indeed not very bright as looking up an empty map is just as efficient as checking whether an option is `None` either way.

Leonidas-from-XIV (approved):
Nice, let's merge it!

PR #309, opened 2022-06-08
Save CLI arguments to the lock file

### LINKED ISSUE #308: Write `opam-monorepo` invocation into lock file
In #307 we discussed how to make the lockfiles reproducible in the presence of command line options (that are able to change the output of the lockfile in a non-trivial way). A compromise that would allow both people to overwrite values via the CLI and yet give hints how to reproduce the generated lockfile is to write the invocation of the `opam-monorepo` call that was used to create the lockfile into the lockfile itself.

The potential advantages are:

1. It allows people who want to reproduce the lock file to reproduce it. At least provided they use the same version of `opam-monorepo` but we don't really provide cross-version reproducibility for plenty of good reasons.
2.  It serves as documentation how the lockfile came to be, so newcomers to a project using `opam-monorepo` are aware of it and due to the invocation written there, they have a pointer to see how to use `opam-monorepo`

While this feature will not be implemented in #307, this issue was created to keep track of the idea.

NathanReb:
Fixes #308

### DIFF
--- a/lib/lockfile.ml
+++ b/lib/lockfile.ml
@@ -261,6 +261,21 @@ module Depexts = struct
     List.concat all |> List.sort_uniq ~cmp:compare_elm
 end
 
+module Cli_args = struct
+  (** Field used to store the raw command line arguments passed to [lock].
+      It is set but not read and therefore is not part of the main lock file
+      type. *)
+
+  type _t = string list
+
+  let name = "cli-args"
+  let shape = Serial_shape.(list string)
+  let from_opam_value value = Serial_shape.from_opam_val shape value
+  let to_opam_value t = Serial_shape.to_opam_val shape t
+  let field = Extra_field.make ~name ~to_opam_value ~from_opam_value
+  let add t opam = match t with [] -> opam | _ -> Extra_field.set field t opam
+end
+
 type t = {
   version : Version.t;
   root_packages : Root_packages.t;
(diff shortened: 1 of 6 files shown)


lib/lockfile.ml
```diff
@@ -261,6 +261,21 @@ module Depexts = struct
     List.concat all |> List.sort_uniq ~cmp:compare_elm
 end
 
+module Cli_args = struct
+  (** Field used to store the raw command line arguments passed to [lock].
+      It is set but not read and therefore is not part of the main lock file
+      type. *)
+
+  type _t = string list
```
Leonidas-from-XIV, line 269: What is the reason to not name it `t`?

lib/lockfile.ml
```diff
@@ -364,11 +379,12 @@ let from_opam ~opam_monorepo_cwd ?file opam =
       source_config;
     }
 
-let save ~opam_monorepo_cwd ~file t =
+let save ~opam_monorepo_cwd ~cli_args ~file t =
   let opam = to_opam ~opam_monorepo_cwd t in
+  let with_cli_args = Cli_args.add cli_args opam in
```
Leonidas-from-XIV, line 384: Wouldn't it be nicer to do something like

```suggestion
  let opam = to_opam ~opam_monorepo_cwd t
    |> Cli_args.add cli_args
  in
```
Then it is clear that `opam` still contains a writeable opam file.

Leonidas-from-XIV (commented):
Very nice. I have some nitpicks/clarifications but I think this is even better than what I envisioned in the original discussion.

lib/lockfile.ml
NathanReb: The compiler rightfully complains it is unused otherwise. I still wanted to keep the type to clarify things in the code.

Leonidas-from-XIV (approved):
:ship: it :) (After placating the formatter)

PR #312, opened 2022-06-09
Upgrade to ocamlformat.0.21.0

NathanReb:


Leonidas-from-XIV (approved):
Looks pretty much like what 0.22 would do.

NathanReb:
Yeah I think there are no or very few formatting changes between 0.21 and 0.22 so most code base are unaffected by upgrading from the former to the latter!

@tmattio will confirm!

PR #314, opened 2022-06-10
Fix a bug where opam-provided CLI options would be ignored

NathanReb:


Leonidas-from-XIV (approved):
Yes, this all looks sensible to me. Thanks!

Issue #284, opened 2022-04-15
Support remote URLs in x-opam-monorepo-repositories


NathanReb:
At the moment only local, fs URLs are supported. This was very helpful for testing purposes and I can imagine it can be helpful to some users as well but for most of them, the feature is not usable yet.

Discussing with opam maintainers we need some basic caching mechanism of our own here. I'll write down the specifics in the issue later. If anyone wants to tackle this please do not hesitate to ping me so I can provide guidance.

PR #317, opened 2022-06-14
Support Git in `x-opam-monorepo-opam-repositories`

Leonidas-from-XIV:
The main issue with this code is that it is difficult to test, since we probably don't want our tests to do IO with the outside world, since it creates issues with sandboxing and dependencies on other services being up and reachable.

I tested it by adding `https://opam.ocaml.org` to the `x-opam-monorepo-repositories` in `vendored.opam` and it indeed outputs a lock file that has e.g. a newer dune than what we expect in the test.

Fixes: #284

### DIFF
--- a/test/bin/explicit-repo.t/run.t
+++ b/test/bin/explicit-repo.t/run.t
@@ -27,7 +27,7 @@ the minimal repository created locally:
   > opam-version: "2.0"
   > EOF
   $ (cd git-repository ; git init ; git add -A ; git commit -m "initial commit") > /dev/null 2>&1
-  $ SHORT_HASH=$(git -C git-repository rev-parse --short HEAD)
+  $ SHORT_HASH1=$(git -C git-repository rev-parse --short HEAD)
 
 To find this package, we need to have the git repository in our
 `x-opam-monorepo-opam-repositories` list:
@@ -51,7 +51,7 @@ Therefore the list of repositories in the lockfile should have the a git repo
 that's the same as the previous git-repository path but end with "#$SHORT_HASH"
 
   $ opam show --just-file --raw -fx-opam-monorepo-opam-repositories ./explicit-repo.opam.locked > locked-repos
-  $ grep -Po ".+(?=$SHORT_HASH)" locked-repos
+  $ grep -Po ".+(?=$SHORT_HASH1)" locked-repos
   git+file://$OPAM_MONOREPO_CWD/git-repository#
 
 We also need to make sure that the git cache gets updated when the repository is updated:
@@ -64,10 +64,15 @@ We also need to make sure that the git cache gets updated when the repository is
   > ]
   > EOF
   > (cd git-repository ; git add -A ; git commit -m "New release") > /dev/null 2>&1
+  $ SHORT_HASH2=$(git -C git-repository rev-parse --short HEAD)
 
 Thus a new version of the package has been released in the git repo. Locking it
-anew should pick this version of the dependency:
+anew should pick this version of the dependency as well as the recording the
+new version of the git opam repository.
 
   $ opam-monorepo lock explicit-repo > /dev/null
   $ opam show --just-file --raw -fdepends ./explicit-repo.opam.locked | grep git-dep
   "git-dep" {= "2.0"}
+  $ opam show --just-file --raw -fx-opam-monorepo-opam-repositories ./explicit-repo.opam.locked > locked-repos
+  $ grep -Po ".+(?=$SHORT_HASH2)" locked-repos
+  git+file://$OPAM_MONOREPO_CWD/git-repository#


NathanReb:
Thanks for looking into this! The main problem with this PR at the moment is that by using `pull_tree` directly, it will use the opam cache and will polute it with repositories.

I think here we need an alternate version of `pull_tree` that does not use the opam source cache to prevent it from growing unreasonnably due to the size of repositories.

I think we should instead have a thin and simple cache layer in `/tmp`. Something similar to what you did except that if the directory already exists we use the repo in there instead of pulling it.

For this to work properly, I think it would be reasonable to only accept remote VCS urls (e.g. no straight tarballs) and to resolve all VCS URLs to a commit hash, similarly to what we do for packages sources already.

What do you think?

NathanReb:
Also we could eventually store archives in that temporary cache instead of the full source tree, similarly to what opam does but that can be done later!

Leonidas-from-XIV:
@NathanReb Sorry, took a while to get back to this but I finally managed to continue on this (mostly testing that things work as desired). I've implemented it to support Git repos specifically.

What it does is the following:

1. Creates a folder, `$TMP_DIR/opam-monorepo`. In my case that was `/tmp/opam-monorepo` but it respects the `TMP_DIR` variable so when tests run this is overridden into a dedicated folder (by, I assume, `dune`)
2. If it encounters a `` `http`` repo that points to `opam.ocaml.org` it rewrites it to the git repository.
3. In this folder it creates `cache` and `repos` folders: the `cache` folder is used by the Opam Git implementation to create a bare repo with the revisions. The `repos` folder is used to check out the repos by MD5 of their URL (to avoid sad surprises with valid characters in file systems)
4. This list of now-`file://` paths is passed to the remaining code which then runs the solver. It also rewrites the URLs of the repos in `Source_config` to include the commit-hash of the checked out repo for reproducability, so the lock file contains specific snapshots of the git URLs. I think that's a pretty neat feature.

There is a small test where it asserts that the checkout works and the lockfile URLs are updated. I didn't add tests for the cache because that is supposed to be functionality of `OpamGit.B`. In my manual testing it did work, the initial lock took a while (we might consider adding a log output) while it was cloning the repository, while a subsequent lock was fast because it used the repo. Testing this in a blackbox is somewhat difficult and I didn't want to depend too much on how `OpamGit.B` implements the cache.

I also tested this on opam-monorepo itself, adding these lines to the `.opam(.template)` file lead it to use the repos and successfully lock:

```
x-opam-monorepo-opam-repositories: [
  "https://opam.ocaml.org"
  "git+https://github.com/dune-universe/opam-overlays.git"
]
```

Leaving out the overlays caused the usual solver failures with missing dune-versions, exactly how it works at the moment when the overlays repo is not added to the switch.

NathanReb (commented):
This looks great! Thanks for your hard work on this!

I'm happy that we can reuse opam's git cache implementation, saves us a lot of troubles.

I have no comments about the code and the tests, it all looks great.

My only concern is how we handle resolution of repo URLs to an exact revision. It seems that we pull the sources from the user provided URL and then look at what revision we got.
I can see two problems with this approach:
1. we might cache based on an ambiguous URL, e.g. `git+https://github.com/ocaml/opam-repository#master` is what gets hashed and therefore I'm not entirely sure whether the current implementation will ever try to fetch it again as long as it is cached, meaning a user might wonder why they're not getting the latest version
2. It mix the two concerns of pulling the sources and making the source config unambiguous.

I'm not exactly sure about `1` as I don't fully understand how the cache works yet so please correct me if I misinterpreted how things were run here. I think either way, this should be fairly easy to add to the test by pushing a new version of a package on the default branch of the local git repo in the cram test.

Leonidas-from-XIV:
1. This is not a problem, because there is a split between the cache folder (always the same) and the repo folder (this gehts hashed). The cache in this case contains a single git repo with a rather specific setup - essentially a bag of commits from all kind of repos it is asked to cache, which then get checked out into the repo folders. The whole setup is a bit non-standard, but seems to work pretty well. Your concern about the cache possibly not being updated with new revisions is however a valid one - the test never verified that updates in the repo are noticed. I added the test and it worked out of the box. Definitely useful to have utilized the OPAM Git cache, looks like the code is pretty solid.
2. I am not sure I follow. To figure out which packages exist I need to pull the repo and likewise, to determine which revision I have I need to have the repository available.

NathanReb:
> I am not sure I follow. To figure out which packages exist I need to pull the repo and likewise, to determine which revision I have I need to have the repository available.


Yeah sorry this was poorly explained.

For inidividual packages which have a git source URL (a pinned package for instance), we  first turn this potentially ambiguous URL into an "exact" one, i.e. one that ends with `#<commit hash>` via calls to `git ls-remote`, that is, without pulling.

What I meant in my second point was that we could do something similar here and separate things in two steps:
1. rewrite the source_config so that repositories git URLs are rewritten to exact URLs
2. pull the repositories described in `source_config` locally

If the opam cache helps us with this I'm also happy with the current implementation. It's slightly harder to figure out this resolution happens with the current code but as long as it works I'm happy with this. Maybe adding a comment or a bit of documentation here would help maintain this code in the future?

test/bin/explicit-repo.t/run.t
```diff
@@ -0,0 +1,73 @@
...
+  > ]
+  > EOF
+  > (cd git-repository ; git add -A ; git commit -m "New release") > /dev/null 2>&1
+
+Thus a new version of the package has been released in the git repo. Locking it
+anew should pick this version of the dependency:
+
+  $ opam-monorepo lock explicit-repo > /dev/null
+  $ opam show --just-file --raw -fdepends ./explicit-repo.opam.locked | grep git-dep
+  "git-dep" {= "2.0"}
```
NathanReb, line 73: This may sound a bit overkill but what would you think about redoing the git hash check here as well? That way we'll test that not only we pull the right version (from the result) but we also deal with opam's cache in the right way to write down the commit hash.

I could think of scenarios where we do the former right but mess up the latter.

NathanReb (approved):

PR #321, opened 2022-07-08
Fix parsing of lock files with empty duniverses

### LINKED ISSUE #319: If all packages are opam-provided, an invalid lockfile is created
Is opam-monorepo so powerful that it can create a lockfile that it can't pull? Turns out, yes :/

```
 $ cat src.opam
opam-version: "2.0"
depends: ["lwt"]
x-opam-monorepo-opam-provided: ["lwt"]
 $ opam monorepo lock        
==> Found 16 opam dependencies for the target package.
==> Wrote lockfile with 0 entries to src.opam.locked. You can now run opam monorepo pull to fetch their sources.
 $ opam monorepo pull
==> Using lockfile src.opam.locked
opam-monorepo: [ERROR] Missing x-opam-monorepo-duniverse-dirs field in opam-monorepo lockfile /tmp/bug/src.opam.locked
```

The reason is that the lockfile does not, indeed, contain this field.

After a light investigation, it seems that it's due to how the opam libs render empty lists:

```
utop # OpamFile.OPAM.empty
  |> OpamFile.OPAM.with_extensions
      (OpamStd.String.Map.singleton "x-field"
        (OpamTypesBase.nullify_pos (OpamParserTypes.FullPos.List (OpamTypesBase.nullify_pos []))))
  |> OpamFile.OPAM.write_to_string;;
- : string = "opam-version: \"2.0\"\n"
```

So from the point of view of `opam-monorepo`, I think that it's necessary to handle the absence of this field as equivalent to an empty list. A special empty value is possible too I guess but since lockfiles are versioned I don't think there can be a confusion between "no field" and "an empty list".

NathanReb:
Fixes #319

### DIFF
--- a/test/bin/empty-duniverse.t/run.t
+++ b/test/bin/empty-duniverse.t/run.t
@@ -1,0 +1,30 @@
+One should be able to generate a lock file where no dependencies need to
+be vendored and that should result in an empty duniverse.
+That can happen if one has no external deps besides 'dune' and 'ocaml'
+or if those are marked as 'opam-provided' for example.
+
+Here our local package only depends on 'ocaml' and 'dune':
+
+  $ opam show --just-file -fdepends ./empty-duniverse.opam
+  ocaml, dune
+
+We should be able to successfully lock:
+
+  $ gen-minimal-repo
+  $ opam-monorepo lock > /dev/null
+
+And the lock file should not contain anything to vendor:
+
+  $ opam show --just-file -fdepends ./empty-duniverse.opam.locked | grep "\?vendor"
+  [1]
+
+Finally, we should be able to pull this lock file, the tool will inform us that
+there is nothing to pull and will not create a duniverse:
+
+  $ opam-monorepo pull
+  ==> Using lockfile $TESTCASE_ROOT/empty-duniverse.opam.locked
+  opam-monorepo: [ERROR] Missing x-opam-monorepo-duniverse-dirs field in opam-monorepo lockfile $TESTCASE_ROOT/empty-duniverse.opam.locked
+  [1]
+  $ ls duniverse
+  ls: cannot access 'duniverse': No such file or directory
+  [2]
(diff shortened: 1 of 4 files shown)


test/bin/empty-duniverse.t/run.t
```diff
@@ -0,0 +1,29 @@
...
+  [1]
+
+Finally, we should be able to pull this lock file, the tool will inform us that
+there is nothing to pull and will not create a duniverse:
+
+  $ opam-monorepo pull
+  ==> Using lockfile $TESTCASE_ROOT/empty-duniverse.opam.locked
+  ==> No dependencies to pull, there's nothing to be done here!
+  $ ls duniverse
+  ls: cannot access 'duniverse': No such file or directory
```
Leonidas-from-XIV, line 28: I am a bit worried that this `ls` output might not be portable. I tried it on Linux with GNU userland and it works, but I am not sure this would be the same on BSD systems.

Maybe use `find . -type d -name 'duniverse'` or something like this? Or maybe `[ -d` to test that `duniverse` is a directory?

Leonidas-from-XIV (commented):
Very nice, the code looks quite elegant. I'm only slightly worried about the fragility of a test, but that's more for ill-defined portability reasons.

test/bin/empty-duniverse.t/run.t
NathanReb: Good point, I could also silence the output and just show that it fails but indeed find seems good too!

Leonidas-from-XIV (approved):
Very nice, let's get it in for the next release!

Leonidas-from-XIV:
Looks like my merge of the bugfix created a changelog conflict, sorry for that.

NathanReb:
My bad, forgot to merge after the review! Just rebased and fixed the conflict, will merge when it goes green!

PR #322, opened 2022-07-12
Disable filtering of available packages for availability of their depexts

Leonidas-from-XIV:
OPAM by default filters out packages for which the system on which it is running does not have the depexts. Which is nice in theory but when building a lockfile we want all packages to be available.

The problem here is that some configurations do not have a queryable package list, e.g. when using Docker containers the package list might have been deleted to save on space in the container, thus `available_packages` would be just a subset and missing packages, despite these existing in the OPAM repositories.

This is the reason why on the Tezos CI it needs `apk update` because if the package list is missing it will just silently exclude a lot of the necessary `conf-*` packages. I tested it locally in a Docker container built in the same way as Tezos where I could reproduce the issue and disabling this in the config makes it work out of the box, without the need for workarounds.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,6 +11,9 @@
 
 ### Fixed
 
+- Enable locking of packages with depexts even with uninitialized system
+  package manager state (#322, @Leonidas-from-XIV)
+
 ### Removed
 
 ### Security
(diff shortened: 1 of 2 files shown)


NathanReb (approved):
Looks good to me! Looking forward to drop this solving method entirely!

Issue #258, opened 2022-02-17
Shouldn't fail when depext fails


Julow:
I understand why the depext tool would fail but I think opam-monorepo shouldn't fail in response.

The actual dependencies are probably already installed and if they aren't, it is not hard to figure out later because the `conf-` packages will tell me.

```
opam-monorepo: internal error, uncaught exception:
               (Failure "External dependency handling not supported for OS family 'nixos'.")
               Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
               Called from CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 31, characters 17-27
               Re-raised at CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 36, characters 4-11
               Called from OpamSysInteract.packages_status in file "duniverse/opam/src/state/opamSysInteract.ml", line 195, characters 8-17
               Called from Duniverse_cli__Depext.run.(fun) in file "cli/depext.ml", line 29, characters 20-56
               Called from OpamGlobalState.with_ in file "duniverse/opam/src/state/opamGlobalState.ml", line 186, characters 14-18
               Re-raised at OpamStd.Exn.finalise in file "duniverse/opam/src/core/opamStd.ml", line 1372, characters 4-38
               Called from Cmdliner_term.app.(fun) in file "duniverse/cmdliner/src/cmdliner_term.ml", line 25, characters 19-24
               Called from Cmdliner.Term.term_result.(fun) in file "duniverse/cmdliner/src/cmdliner.ml", line 33, characters 27-34
               Called from Cmdliner.Term.run in file "duniverse/cmdliner/src/cmdliner.ml", line 117, characters 32-39
```

Julow:
I just noticed that `opam-monorepo depext` was explicitly called so failing is justified. I somehow thought that `opam-monorepo pull` was failing.

Still the error message could be improved.

NathanReb:
Thanks for reporting this!

Yeah we should definitely catch that and report it properly.

PR #323, opened 2022-07-27
Properly display error message when depext fails

RyanGibb:
Addresses #258

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -15,6 +15,8 @@
   package manager state (#322, @Leonidas-from-XIV)
 - Fix a bug where `pull` would crash if the lock file contained no package to
   vendor (#321, @NathanReb)
+- Display a better error message when the depext command fails when getting the
+  status of the packages (#258, #323, @RyanGibb, @Julow)
 
 ### Removed
 
--- a/cli/depext.ml
+++ b/cli/depext.ml
@@ -43,7 +43,7 @@ let run (`Root root) (`Lockfile explicit_lockfile) dry_run (`Yes yes) () =
                 Ok ()
               with Failure msg -> Error (`Msg msg)
             else Ok ()
-        with Failure msg -> Error (`Msg msg))
+      with Failure msg -> Error (`Msg msg))
 
 open Cmdliner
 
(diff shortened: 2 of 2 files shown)


RyanGibb:
Before:
```
               Failure("External dependency handling not supported for OS family 'nixos'.")
               Called from CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 31, characters 17-27
               Re-raised at CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 36, characters 4-11
               Called from OpamSysInteract.packages_status in file "duniverse/opam/src/state/opamSysInteract.ml", line 224, characters 8-17
               Called from Duniverse_cli__Depext.run.(fun) in file "cli/depext.ml", line 29, characters 20-56
               Called from OpamGlobalState.with_ in file "duniverse/opam/src/state/opamGlobalState.ml", line 186, characters 14-18
               Re-raised at OpamStd.Exn.finalise in file "duniverse/opam/src/core/opamStd.ml", line 1372, characters 4-38
```

After:
```
opam-monorepo: [ERROR] External dependency handling not supported for OS family 'nixos'.
```

Leonidas-from-XIV:
Hi @RyanGibb, thanks for your PR! This is a pretty large `try` block, could you rewrite it in a way where it exits directly on the error and only continues if no exception is thrown? It looks like this could be lifted into the `Error` monad, that would make reading the code easier since it would avoid nested `try` blocks and pin-point the location of the failure (`OpamSysInteract.packages_status`) directly.

RyanGibb:
Hi @Leonidas-from-XIV, thanks for the feedback! Yes of course. Is this a little better?

NathanReb (approved):
Looks good to me, thanks for the fix!

@Leonidas-from-XIV we might want to add an integration test for this, I wonder if we can easily write a portable one.

Leonidas-from-XIV:
@NathanReb Both #258 as well as this PR mention `nixos` so I looked at how the failure happens and this is is when OPAM tries to determine the OS family. It does so by determining the OS and only if it is Linux it goes ahead and checks `/etc/os-release`. So we would need to fake `/etc/os-release` and fake `uname -s` output (for it to also fail on macOS) which seems quite complicated to do portably. Integration tests around OPAM are unfortunately quite a pain :(

PR #324, opened 2022-08-01
Update lock file

NathanReb:
Nothing significant here, not even 5.0.0 compat as the packages required for this (base and ocaml) are still marked as avoid versions for now.

If the need arise we can force their selection by locking again with `--ocaml-version 5.0.0`,


Leonidas-from-XIV (approved):
I was confused why so many lines disappear but it seems like `bigarray-compat` and `mmap` disappeared from our dependency cone as no longer needed transitive dependencies. That's a welcome change.

PR #325, opened 2022-08-02
Update the lock file to OCaml 5

NathanReb:
I'm creating this as a draft PR to showcase OCaml 5 compat.

The main and only blocker is `base` which has an `avoid-version` flagged release compatible with OCaml 5. Main problem is that this version claims to be compatible with OCaml 5 only so there's no guarantee we wouldn't lose compat with older compilers.

I did test this with 4.14 and it worked so the OCaml bounds on `base.v0.15.0~5.0preview` are too strict, which is encouraging.

We could, in theory, merge this already but I'd much rather wait until it does not depend on `avoid` flagged versions (outside of OCaml) first! The only way we can get this lock file atm is by explicitly asking for `ocaml.5.0.0`.

### DIFF
--- a/opam-monorepo.opam.locked
+++ b/opam-monorepo.opam.locked
@@ -6,9 +6,11 @@ depends: [
   "alcotest" {= "1.6.0" & ?vendor}
   "angstrom" {= "0.15.0" & ?vendor}
   "astring" {= "0.8.5+dune" & ?vendor}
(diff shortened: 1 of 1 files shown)


NathanReb:
It even builds all the way back to 4.10!

Leonidas-from-XIV:
Obsoleted by #350, now that `base.v0.15.1` was released.

Issue #327, opened 2022-08-13
Using dkml-base-compiler instead of ocaml-base-compiler


jonahbeckford:
[dkml-base-compiler](https://v3.ocaml.org/p/dkml-base-compiler/4.12.1~v1.0.0) is a new compiler package in the Opam repository. Among other things, it builds a cross-compiler for macOS (x86_64 + ARM64) and unofficially can build cross-compilers for Android (ARM32, etc.).

I can't figure out the command line options so that `dkml-base-compiler.4.12.1~v1.0.0` is used as the compiler during `opam monorepo lock` rather than `ocaml-base-compiler.VERSION`.

Context: For my own internal needs I have been using the dunified Opam repository but do not use the `opam monorepo` tools; instead I use Opam build command configuration so that `opam install .` uses `dune -x darwin_arm64` to build every Opam package dependency. That style of cross-compilation may be too complicated for other people. So not only would I like others to use `opam monorepo` with `dkml-base-compiler` but I've got [some "dkml-workflows" tooling I'll be releasing](https://github.com/diskuv/dkml-workflows-monorepo-example) that will create "native" (including cross-compiled) artifacts on GitHub Actions.

If needed, I have a [Makefile](https://github.com/diskuv/dkml-workflows-monorepo-example/blob/74fde8bdd65b7fa653a9a906876307a62f7170d1/Makefile) that can be used to see the problems doing `opam monorepo lock` with `dkml-base-compiler`:

```console
$ git clone https://github.com/diskuv/dkml-workflows-monorepo-example.git
$ cd dkml-workflows-monorepo-example
$ git reset --hard 2afd44bb59d41cf38cb382ad679b03b16fbffc85
$ make monorepo-pull
opam switch create . dkml-base-compiler.4.12.1~v1.0.0 \
          --yes \
          --deps-only \
          --repo default=https://opam.ocaml.org,dune-universe=git+https://github.com/dune-universe/opam-overlays.git

<><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><>  🐫 
Switch invariant: ["dkml-base-compiler" {= "4.12.1~v1.0.0"}]
The following actions will be performed:
  ∗ install dkml-base-compiler        4.12.1~v1.0.0
  ∗ install base-unix                 base
...
Done.
# Run eval $(opam env) to update the current shell environment
[...]

PR #328, opened 2022-08-15
Read package flags from OPAM metadata

Leonidas-from-XIV:
This changes the logic from hardcoding the compiler "base packages" to instead read the OPAM flags to check whether the package in question is a compiler in which case it will be considered a base package.

It also adds code to check for packages whether they are `conf` packages, which could be used to check for virtual packages.

Inspired by #327, this would allow `dkml-base-compiler` to be considered a base package and not accidentally vendor it. However it doesn't yet allow using the DKML compiler, because the name of the `ocaml` conf package is still hardcoded with no way to override it (yet).

### DIFF
--- a/lib/opam.ml
+++ b/lib/opam.ml
@@ -230,14 +230,14 @@ module Package_summary = struct
     | { dev_repo = None | Some ""; _ } -> true
     | _ -> false
 
-  let is_base_package v =
-    let in_base_pkgs { package; _ } =
-      OpamPackage.Name.Set.mem package.name Config.base_packages
-    in
-    let is_compiler_pkg { package; _ } =
-      OpamPackage.Name.equal package.name Config.compiler_package_name
-    in
-    is_compiler v || in_base_pkgs v || is_compiler_pkg v
+  let is_compiler_package { package; _ } =
+    OpamPackage.Name.equal package.name Config.compiler_package_name
+
+  let is_base_package { package; _ } =
+    OpamPackage.Name.Set.mem package.name Config.base_packages
+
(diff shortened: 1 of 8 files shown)


lib/opam.ml
```diff
@@ -217,18 +245,25 @@ module Package_summary = struct
...
     | { dev_repo = None | Some ""; _ } -> true
     | _ -> false
 
-  let is_base_package = function
-    | { package; _ }
-      when OpamPackage.Name.Set.mem package.name Config.base_packages ->
-        true
-    | _ -> false
+  let is_base_package v =
+    let in_base_pkgs { package; _ } =
```
gridbugs, line 260: I find this naming confusing as "base package" now means both a package appearing in the list of base packages in the config (`in_base_pkgs`) and a package which appears in the base packages list or is a compiler packages (`is_base_package`).

Leonidas-from-XIV: That's a good point. I've looked into this a bit and realized that all the checks always test for `is_base` and `is_virtual` at the same time so I created `is_safe` that calls `is_base`, `is_virtual` and also the new flag-based variants.

`is_safe_package` is not a great name, better suggestions welcome.

Leonidas-from-XIV:
@gridbugs Can you take a look again? I've done some renaming and while thinking about it realized that even to me it was unclear what constituted a "base" package. With the compiler gone from the list it turns out that these are mostly build systems and dependencies of `oasis` (that aren't much used outside of it) so I explained the reasoning for them.

gridbugs (approved):

PR #332, opened 2022-08-30
Allow resolving unprefixed refs

### LINKED ISSUE #326: failure with a pinned opam package
I try to use opam monorepo 0.3.3 with the following pins:
`gmp.6.2.1-2  (uninstalled)  git  git+file:///usr/home/hannes/devel/mirage/ocaml-gmp#HEAD`

And receive a failure during `opam monorepo lock`:
```
gmake[2]: Leaving directory '/usr/home/hannes/devel/mirage/mirage-skeleton/applications/static_website_tls'
 ↳ generate lockfile for monorepo dependencies
==> Found 130 opam dependencies for the target package.
opam-monorepo: [ERROR] No HEAD ref for file:///usr/home/hannes/devel/mirage/ocaml-gmp
```

It would be pretty nice to support pins to git repositories the same as opam does, or are there any blockers?

Leonidas-from-XIV:
When using a pinned dependency the refs might not have a prefix (e.g. `HEAD` is a plain ref) thus they wouldn't have been found when attempting to resolve the refs.

Closes #326

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,9 @@
 
 ### Fixed
 
+- Fix resolving refs of locally pinned repositories (#326, #332, @hannesm,
+  @Leonidas-from-XIV)
+
 ### Removed
 
 ### Security
--- a/lib/git.ml
+++ b/lib/git.ml
@@ -56,12 +56,19 @@ module Ls_remote = struct
           let result = search_ref (prefix ^ ref) parsed_lines in
           interpret_search_result result
         in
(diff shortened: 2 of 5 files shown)


Leonidas-from-XIV:
@samoht I've added a test. It's a bit ugly since it messes with the global OPAM state (by creating a pin called `opam-monorepo-test-other` but the original report can only be reproduced with pins.

gridbugs (approved):

PR #338, opened 2022-10-22
Fix use of x-opam-monorepo-global-opam-vars in docs

patricoferris:
This took a bit of head scratching to understand why this field wasn't working locally ^^"


Leonidas-from-XIV:
Thanks! It even took me a moment to figure out what changed in this PR :sweat_smile:

PR #344, opened 2022-10-31
Respect the "depext" field of global opam config

gridbugs:
Prior to this change, opam-monorepo will attempt to install external dependencies, even if this field is set to false.


gridbugs:
The context for this change is that I was trying to do the mirage tutorial on nixos and was running into an error when opam-monorepo tried to check the status of depexts:
```
               Failure("External dependency handling not supported for OS family 'nixos'.")
               Called from CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 31, characters 17-27
               Re-raised at CamlinternalLazy.force_lazy_block in file "camlinternalLazy.ml", line 36, characters 4-11
               Called from OpamSysInteract.packages_status in file "duniverse/opam/src/state/opamSysInteract.ml", line 224, characters 8-17
               Called from Duniverse_cli__Depext.run.(fun) in file "cli/depext.ml", line 29, characters 20-56
               Called from OpamGlobalState.with_ in file "duniverse/opam/src/state/opamGlobalState.ml", line 186, characters 14-18
               Re-raised at OpamStd.Exn.finalise in file "duniverse/opam/src/core/opamStd.ml", line 1372, characters 4-38
```

Opam doesn't know how to install depexts on nixos, and even if it did, I'd rather it not mess with my system packages.

Possibly relevant to https://github.com/tarides/opam-monorepo/issues/258

Leonidas-from-XIV:
Thanks, could you also add a command line flag to control the setting? As-is it makes builds less reproducible as it reads more from opam settings (a thing that we're gradually trying to reduce to zero for higher reproduceability, with all settings controlled by the CLI/opam-files).

Leonidas-from-XIV:
(I am confused why even despite #323 you get this stacktrace, there is some kind of error still lurking in the code)

gridbugs:
> (I am confused why even despite #323 you get this stacktrace, there is some kind of error still lurking in the code)

The stacktrace only prints out when I run opam-monorepo as an opam plugin. If I run opam-monorepo instead, the output is:
```
==> Using lockfile mirage/noop-unix.opam.locked
opam_monorepo.exe: [ERROR] External dependency handling not supported for OS family 'nixos'.
```

gridbugs:
> Thanks, could you also add a command line flag to control the setting? As-is it makes builds less reproducible as it reads more from opam settings (a thing that we're gradually trying to reduce to zero for higher reproduceability, with all settings controlled by the CLI/opam-files).

I added a flag `--disable-depext` which causes the same behaviour as setting `depext: false` in the global config.

gridbugs:
> I would suggest to make the flag `--resolve-depext=true|false|config` (or `yes|no|auto`, idk) with the default of `config`, that way it can be forced either way, without having to add more flags like `--force-depext`.

Good suggestion. Updated!

cli/depext.ml
```diff
@@ -31,21 +32,41 @@ let run (`Root root) (`Lockfile explicit_lockfile) dry_run (`Yes yes) () =
...
+            else Ok ()
+      else (
+        Logs.warn (fun l ->
+            let message =
+              "Refusing to check for external dependencies as depext is \
+               disabled.\n\
+               The following packages may need to be installed manually:"
+            in
+            l "%s\n%s" message
+              (String.concat ~sep:"\n"
```
Leonidas-from-XIV, line 65: You can use `Fmt.list ~sep:(Fmt.any "\n") string` to avoid that concat.

Or even (untested, but I enjoy how much one can do with `Fmt` combinators):
```ocaml
let pp_sys_pkg = Fmt.of_to_string OpamSysPkg.to_string in
let pp_item ppf = Fmt.pf ppf "- %a" pp_sys_pkg in
l "%s\n%a" message (Fmt.list ~sep:(Fmt.any "\n") pp_item) (OpamSysPkg.Set.elements pkgs)
```

Leonidas-from-XIV (changes requested):
Yes, this is nice. Some small comments on how to make it more consistent/readable but the concept is solid now.

Leonidas-from-XIV:
I think @RyanGibb mentions a good point [here](https://github.com/ocaml/opam/pull/5332#issuecomment-1299986626). Maybe it makes no sense to call `opam-monorepo depext` if the depexts are not supposed to be computed? From a user perspective it might be rather strange to call `opam-monorepo depext` and then nothing happens because it was disabled in the configuration?

PR #345, opened 2022-11-10
Fix failed merge of cache URL handling

Leonidas-from-XIV:
Merging #337 lead to a case where the merging didn't yield a working solution.

This PR fixes it by changing the code to:

- Retrieve the cache urls from the `global_state` as #337 does
- Do not pass cache urls if we're not dealing with an OPAM state


samoht:
Merging to fix the build, but some tests are still broken.

PR #350, opened 2022-11-11
Update the lock file to OCaml 5

Leonidas-from-XIV:
Now that `base.v0.15.1` exists, we can update the locked dependencies to "final" releases.

This obsoletes https://github.com/tarides/opam-monorepo/pull/325


tmattio:
Should we merge this? I think it would allow to cut a release of opam-monorepo compatible with OCaml 5.

Leonidas-from-XIV:
Yes, definitely!

tmattio:
cc @gridbugs do you think you'd have time to give it a quick look?

samoht:
LGTM

Issue #145, opened 2020-12-15
Inconsistent solution when incompatible packages from the same repository are selected


emillon:
Hi,

When requesting different versions from packages that come from the same repository, some invalid solutions can be selected. For example requesting dune 2.7.1 and dune-build-info 2.6.2 can create the following lockfile fragment:

```
  [
    "dune-build-info.2.6.2"
    "https://github.com/ocaml/dune/releases/download/2.7.1/dune-2.7.1.tbz"
  ]
```

NathanReb:
This is intended behaviour atm. The only clean alternative is to reject such solutions which in some cases mean there won't be any solutions.

One thing I considered was to re-run the solver by adding constraints so that all multi-opam repo packages share the same version but I'm not even sure we can express that constraint and there would still be cases where there is no such solution (e.g. one of the packages hasn't been released together with the others).

The ideal would be to have per package source tree pruning in dune, meaning we could pull packages from multi-opam repos separately which would completely rule out the need to find a compatible version for all those packages.

In the meantime we have to select a version for the tool to remain usable so we pick the highest one.

I agree this is not satisfying but unfortunately I don't think there is a simple solution on our end. In particular note that this is very fragile as if one was to version packages differently in the same repo, the version selected might not be the most recent one according to the repo's VCS history.

I'll open an issue on dune to see if pruning would be an option we're willing to consider.

emillon:
I think it would be OK to just fail in that case, or at maybe just emit a warning.

NathanReb:
I emphasize that failing is definitely not an option here unfortunately as it would fail pretty much all the time. From experience, those are very common.

From a user perspective, the only solution would be to require a fixed version for all such packages which is not satisfying at all and again, not always valid!

NathanReb:
A warning would be good indeed, the log already exist but it's `info` level atm, meaning it only shows up with `-v`!

emillon:
Raising the level seems good enough in the short term.

In the medium term I'd be curious to see if adding the extra constraints can make it work. These constraints reflect the reality (in the opam-monorepo model, the same version is used so there are extra constraints compared to the opam model).

NathanReb:
> In the medium term I'd be curious to see if adding the extra constraints can make it work. These constraints reflect the reality (in the opam-monorepo model, the same version is used so there are extra constraints compared to the opam model).

Indeed they reflect the reality in the cases where the selected version exists for all packages in the repo but that's not always true.

[...]

PR #353, opened 2022-11-18
WIP: Respect pins

reynir:
This is an attempt at improving the situation described in #145 by always using the source of pinned packages when there's a 'collision'. If there are multiple packages with the same repository pinned to different URIs opam-monorepo fails.

- [ ] Fail more gracefully than raising `Failure _`.
- [ ] Tracking pinned status could probably be done more gracefully.

This improves for me the situation described in https://github.com/mirage/mirage/issues/1375

### DIFF
--- a/lib/duniverse.ml
+++ b/lib/duniverse.ml
@@ -58,19 +58,22 @@ module Repo = struct
       dev_repo : string;
       url : unresolved Url.t;
       hashes : OpamHash.t list;
+      pinned : bool;
     }
 
     let equal t t' =
       OpamPackage.equal t.opam t'.opam
       && String.equal t.dev_repo t'.dev_repo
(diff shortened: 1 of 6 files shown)


reynir:
Hello! I think this is a very useful change. It's soon been half a year without a comment which has been very unmotivating for me. Is this a wont-fix then?

samoht:
Does this work as well when packages are pinned with `opam pin <pkg> --dev` in the current switch. Last time I checked, the logic to detect those packages was very ad-hoc (the package has a single version, and the URL doesn't have a checksum) and I don't see this little dance happening here.

reynir:
I honestly don't remember what this PR does. I found `OpamSwitchState.is_pinned` which can check, given a package name, if a package is pinned.

reynir:
Superseded by #398 (thanks @samoht!)

PR #354, opened 2022-11-22
Dune 3.5 update

Leonidas-from-XIV:
This is useful for releases where we're supposed to bump the dune-lang version. But opam-monorepo is currently not compatible with newer versions of dune so this PR attempts to fix things.

### DIFF
--- a/cli/common.ml
+++ b/cli/common.ml
@@ -108,11 +108,11 @@ module Logs = struct
   let app ?src f =
     Logs.app ?src (fun l ->
         f (fun ?header ?tags fmt ->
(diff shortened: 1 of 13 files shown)


Leonidas-from-XIV:
This PR has been affected by https://github.com/ocaml/dune/issues/6545 so there's a bit of code reshuffling to not accidentally refer to `Duniverse_lib.Duniverse_lib` (which is `struct end` in dune 3.5 and 3.6.0) and while there is already a PR with a fix I think cleaning up the code no not accidentally refer to it is a good practice anyway.

gridbugs (approved):

PR #355, opened 2022-11-23
Allow packages with no build commands

gridbugs:
If a package has no build commands, chances are that it's a metapackage which exists just so that its dependencies can be installed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 
 ### Changed
 
+- Treat packages with no build commands as if they can be built with dune (#355,
+  @gridbugs)
+
 ### Deprecated
 
 ### Fixed
--- a/lib/opam.ml
+++ b/lib/opam.ml
@@ -129,6 +129,14 @@ module Hash = struct
     Format.fprintf fmt "%s:%s" kind contents
 end
 
+module Command = struct
+  type t = OpamTypes.command
+
+  let pp fmt (t : t) =
+    OpamPp.print OpamFormat.V.command t
+    |> OpamPrinter.FullPos.value |> Format.pp_print_string fmt
+end
+
 module Depexts = struct
   let pp fmt t =
     let pp_filter fmt filter =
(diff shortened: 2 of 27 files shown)


lib/opam_solve.ml
```diff
@@ -95,11 +95,12 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
       Opam.depends_on_dune ~allow_jbuilder depends
       || Opam.depends_on_dune ~allow_jbuilder depopts
     in
+    let has_no_build_command = List.is_empty (OpamFile.OPAM.build opam_file) in
     let summary = Opam.Package_summary.from_opam pkg opam_file in
     let is_valid_dune_wise =
       Opam.Package_summary.is_base_package summary
       || Opam.Package_summary.is_virtual summary
```
Leonidas-from-XIV, line 102: I'm wondering, wouldn't it make sense to add this to `is_virtual`? Or a new category in `Opam.Package_summary`?

Leonidas-from-XIV (commented):
It's a useful change, it sounds like packages that don't need to be "built" can be safely included into the duniverse.

Could you add a test with a package that has no build instructions?

lib/opam_solve.ml
gridbugs: I considered that and conceptually it would make sense for packages without build commands to be considered "virtual". We could add a new field to `Package_summary.t`:
```ocaml
build : OpamTypes.command list
```
The problem is that in order to update `Package_summary.equal` and `Package_summary.pp` we'd need to implement equality and pretty-printing for `OpamTypes.command` and it looks like that would add a lot of code.

However, it appears that neither `Package_summary.equal` nor 'Package_summary.pp' are ever called, so perhaps they can just be deleted.

gridbugs: I've updated this PR so that the check is now done inside `is_virtual`

Leonidas-from-XIV (commented):
Looks very good! Just make the code a bit less brittle and then this is ready to merge :ship:

Leonidas-from-XIV (approved):
Nice, it's ready to go. Thanks for the PR!

PR #356, opened 2022-11-24
Remove some dead code

gridbugs:

In https://github.com/tarides/opam-monorepo/pull/355 I wanted to add a new field to `Package_summary.t`:
```ocaml
build : OpamTypes.command list
```
...but found it would be non-trivial to implement `equal` or `pp` for the `OpamTypes.command` type. This would only be necessary to update the `Package_summary.equal` and `Package_summary.pp` functions, but I found that they aren't called anywhere in this project.

### DIFF
--- a/lib/opam.ml
+++ b/lib/opam.ml
@@ -130,12 +130,6 @@ module Hash = struct
 end
 
 module Depexts = struct
-  let equal t t' =
-    List.equal
-      (fun (pkg_set, filter) (pkg_set', filter') ->
-        OpamSysPkg.Set.equal pkg_set pkg_set' && filter = filter')
-      t t'
-
   let pp fmt t =
     let pp_filter fmt filter =
       Format.fprintf fmt "%s" (OpamFilter.to_string filter)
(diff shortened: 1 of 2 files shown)


lib/opam.ml
```diff
@@ -189,24 +170,6 @@ module Package_summary = struct
...
   }
 
-  let equal { package; url_src; hashes; dev_repo; depexts } t' =
-    OpamPackage.equal package t'.package
-    && Option.equal Url.equal url_src t'.url_src
-    && List.equal Hash.equal hashes t'.hashes
-    && Option.equal String.equal dev_repo t'.dev_repo
-    && Depexts.equal depexts t'.depexts
-
-  let pp fmt { package; url_src; hashes; dev_repo; depexts } =
```
Leonidas-from-XIV, line 199: I would keep `pp` because… I often use it for print debugging and it is convenient to be able to do a debug print every now and then :see_no_evil:

CHANGES.md
```diff
@@ -10,6 +10,8 @@
 
 ### Removed
 
+- Remove unused `Package_summary.equal` and `Package_summary.pp` functions (#356, @gridbugs)
```
Leonidas-from-XIV, line 13: This doesn't need a changelog entry I would say since it's not user-visible (at least if the user doesn't use `Duniverse_lib` but I'd argue that the user facing side of opam-monorepo is the binary, not the library).

I've tagged it as "no changelog" so the CI will pass without requiring an entry.

Leonidas-from-XIV (commented):
Always great to remove dead code, especially stuff that the compiler is not warning about.

I've added a few comments, since I kinda like `pp` but the rest is good to go!

lib/opam.ml
gridbugs: Fair enough. I've added back `pp`. Looks like implementing `pp` for build commands won't be too hard.

Leonidas-from-XIV (approved):

PR #357, opened 2022-11-28
Make tests compatible with macos

gridbugs:
Some commands used in tests use features not available on macos:
 - `sed -i ...` requires a file extension to be passed as an argument, and the original file is preserved as a copy with the given extension appended
 - `grep -P ...` is not implemented

I updated the tests to use perl instead as its behaviour will be consistent across different unixes.

Also, when using homebrew as a package manager, opam incorrectly reports non-existent packages as available, presumably because brew provides no fast way to query for the existence of non-installed packages. This was causing a depext test to fail on macos.

### DIFF
--- a/test/bin/config-cli-args.t/run.t
+++ b/test/bin/config-cli-args.t/run.t
@@ -47,7 +47,7 @@ Once again, we can see from the generated lock file that it used both repos.
 Note that you can use the '--opam-repositories' option to overwrite any local
 extension.
 
-  $ sed -i '$d' ./config-cli-args.opam
+  $ sed -i -e '$d' ./config-cli-args.opam
   $ echo 'x-opam-monorepo-opam-repositories: ["https://a.com/a.tbz"]' >> ./config-cli-args.opam
   $ opam show --just-file -fx-opam-monorepo-opam-repositories ./config-cli-args.opam
   https://a.com/a.tbz
--- a/test/bin/dune
+++ b/test/bin/dune
@@ -1,6 +1,9 @@
 (cram
  (applies_to :whole_subtree)
- (deps %{bin:opam-monorepo} %{bin:gen-minimal-repo}))
+ (deps
+  %{bin:opam-monorepo}
+  %{bin:gen-minimal-repo}
+  (sandbox preserve_file_kind)))
 
 (executable
  (name gen_minimal_repo)
--- a/test/bin/explicit-repo.t/run.t
+++ b/test/bin/explicit-repo.t/run.t
@@ -51,7 +51,7 @@ Therefore the list of repositories in the lockfile should have the a git repo
 that's the same as the previous git-repository path but end with "#$SHORT_HASH"
 
   $ opam show --just-file --raw -fx-opam-monorepo-opam-repositories ./explicit-repo.opam.locked > locked-repos
-  $ grep -Po ".+(?=$SHORT_HASH1)" locked-repos
+  $ sed "s/\(.*\)$SHORT_HASH1.*/\1/" locked-repos
   git+file://$OPAM_MONOREPO_CWD/git-repository#
 
 We also need to make sure that the git cache gets updated when the repository is updated:
@@ -74,5 +74,5 @@ new version of the git opam repository.
   $ opam show --just-file --raw -fdepends ./explicit-repo.opam.locked | grep git-dep
   "git-dep" {= "2.0"}
   $ opam show --just-file --raw -fx-opam-monorepo-opam-repositories ./explicit-repo.opam.locked > locked-repos
-  $ grep -Po ".+(?=$SHORT_HASH2)" locked-repos
+  $ sed "s/\(.*\)$SHORT_HASH2.*/\1/" locked-repos
   git+file://$OPAM_MONOREPO_CWD/git-repository#
--- a/test/bin/invalid-package-version.t/run.t
+++ b/test/bin/invalid-package-version.t/run.t
@@ -41,8 +41,9 @@ to the test)
   $ opam-monorepo lock toonew 2> errors
   ==> Using 1 locally scanned package as the target.
   [1]
-  $ grep -Pazo "(?s)opam-monorepo: \[ERROR\].*(?=opam-monorepo)" < errors | head --bytes=-1
+  $ grep "opam-monorepo: \[ERROR\]" < errors
   opam-monorepo: [ERROR] There is no eligible version of a that matches >= 1.0
+  opam-monorepo: [ERROR] Can't find all required versions.
 
 We should also produce the right error message with all the constraints when we have multiple constaints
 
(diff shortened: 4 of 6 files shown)


test/bin/config-cli-args.t/run.t
```diff
@@ -47,7 +47,7 @@ Once again, we can see from the generated lock file that it used both repos.
 Note that you can use the '--opam-repositories' option to overwrite any local
 extension.
 
-  $ sed -i '$d' ./config-cli-args.opam
+  $ perl -i -ne 'print unless eof' ./config-cli-args.opam
```
Leonidas-from-XIV, line 50: I don't think depending on `perl` is a good idea here. `perl` is usually not part of POSIX and as such not installed by default on some Linux distributions. We would need to add a dependency on `conf-perl` and I assume Windows users wouldn't be happy either.

Is the `sed -i''` trick not working?

test/bin/explicit-repo.t/run.t
```diff
@@ -51,7 +51,7 @@ Therefore the list of repositories in the lockfile should have the a git repo
 that's the same as the previous git-repository path but end with "#$SHORT_HASH"
 
   $ opam show --just-file --raw -fx-opam-monorepo-opam-repositories ./explicit-repo.opam.locked > locked-repos
-  $ grep -Po ".+(?=$SHORT_HASH1)" locked-repos
+  $ perl -ne "print \$1 if /(.+)$SHORT_HASH1/" locked-repos
```
Leonidas-from-XIV, line 54: If we need `perl` here, I would suggest writing our own clean-up processor in OCaml instead of depending on a fairly large interpreter.

test/bin/depext.t/run.t
```diff
@@ -22,7 +22,11 @@ We lock and expect the OS package to be part of the locked Opam file.
...
 Then we want to make sure `depext` works. Given the fantasy package does not
-exist, `depext` should not install any packages but still work successfully:
+exist, `depext` should not install any packages but still work successfully.
+On some systems, opam reports non-existent packages as available, and such
+packages will appear in the output of `opam-monorepo depext --dry-run`. Filter
+lines containing "libfantasydependency" from the output for consistency across
+systems.
 
-  $ opam-monorepo depext --dry-run
+  $ opam-monorepo depext --dry-run | grep -v libfantasydependency
```
Leonidas-from-XIV, line 31: I've seen this test fail but I fear there's something else going on because it didn't use to fail on my system but since recently it does. It seems the behaviour of OPAM changed at some point.

If my backtracking is correct, this changed between OPAM 2.1.0 and 2.1.1, so we should probably make `opam-monorepo depext` work consistently either on all supported versions of OPAM or raise the dependency and add the behaviour change in a PR (with a changelog entry since it changes user-visible behavior).

test/bin/invalid-package-version.t/run.t
```diff
@@ -41,7 +41,7 @@ to the test)
   $ opam-monorepo lock toonew 2> errors
   ==> Using 1 locally scanned package as the target.
   [1]
-  $ grep -Pazo "(?s)opam-monorepo: \[ERROR\].*(?=opam-monorepo)" < errors | head --bytes=-1
+  $ perl -0ne 'print $1 if /(?s)(opam-monorepo: \[ERROR\].*)opam-monorepo/s' < errors | head --bytes=-1
```
Leonidas-from-XIV, line 44: Probably would be nicer to have a separate program process this, since it's becoming increasingly difficult to understand what is being processed here (and I say that having written the initial line which, a few month down the line looks also puzzling)

Leonidas-from-XIV (changes requested):
Fixing the tests is great. Some of them are getting increasingly ridiculous to understand, with all kinds of jumping-through-hoops to force grep to do what we want it to do. On the other hand, I don't think depending on `perl` is the right way here. In the end we probably need to bite the bullet and write some OCaml helper program to clean up stdout. This should probably also make it easier to understand since ther's no need to unserstand silly options like `grep -Pazo` just to force grep to do the thing we want.

The `libfantasydependency` problem is not related to homebrew, it also happens on Fedora Linux. This looks like a change in OPAM itself that we should probably handle to avoid user breakage.

test/bin/depext.t/run.t
gridbugs: OPAM separates depexts into available, unavailable, and installed. For some package managers, all non-installed packages are assumed to be available. Homebrew is one such package manager. It looks like we should be able to quickly query the brew package repo for available packages using a regexp, similar to how we do it for apt. I'll make a PR to opam to add this.

test/bin/config-cli-args.t/run.t
gridbugs: I got it working by passing the `-e` flag explicitly to `sed`

test/bin/invalid-package-version.t/run.t
gridbugs: I simplified the regex so it matches without the perl regexp extension to grep (the `-P`).

test/bin/dune
```diff
@@ -1,6 +1,6 @@
 (cram
  (applies_to :whole_subtree)
- (deps %{bin:opam-monorepo} %{bin:gen-minimal-repo}))
+ (deps %{bin:opam-monorepo} %{bin:gen-minimal-repo} (sandbox preserve_file_kind)))
```
gridbugs, line 3: Heads up that I added this sandbox configuration for tests. Some of the tests use `sed -i` to modify files in place and this fails when the files are symlinked into the sandbox (fortunately!) which was happening on my mac but not on linux.

Leonidas-from-XIV: Didn't know about it, good catch! I was confused why that worked so I just tested it and it looks like GNU sed just recreates the "inplace" file as a regular file instead of following the symlink.

test/bin/depext.t/run.t
gridbugs: Here's the PR that lets opam check for available packages with homebrew: https://github.com/ocaml/opam/pull/5372

Leonidas-from-XIV: Is that separation something that has been added in 2.1.1? Because on my Fedora Linux at hand the behavior of `opam-monorepo` changes whether using opam-core 2.1.0 or 2.1.1, so it seems like the availability check was removed in that case which I think is would be odd.

test/bin/explicit-repo.t/run.t
gridbugs: I've changed the tests back to use a combination of sed and grep rather than perl, but using the common features available in the gnu and bsd versions so they work on linux and macos.

test/bin/depext.t/run.t
gridbugs: I did some archeology and it looks like support for checking for available packages on fedora and friends was remove intentionally between 2.1.0 and 2.1.1 in this commit https://github.com/ocaml/opam/commit/c76a512a4d887a6e9623c6f5447ff8d26104cc49. It's unrelated to the lack of support for checking for available packages on macos with homebrew.

Leonidas-from-XIV: Ah, ok. I see. Thanks for the investigation. I feared this would be a general change in OPAM and while I am not super happy about this happening in a patch version I guess there's not much worth doing on our side.

Leonidas-from-XIV (approved):
Looks good! Unfortunately the CI doesn't test on macOS (we should consider submitting additional features to OCaml-CI where it ignores the lockfile and builds the packages normally since there is a macOS target now) but I assume the tests pass on your machine :)

PR #359, opened 2022-11-28
Fix bug where dev repo urls ending with a "/" would result in  `opam monorepo pull` placing package source code directly inside the duniverse  directory instead of in a subdirectory of the duniverse directory

gridbugs:


lib/dev_repo.mli
```diff
@@ -5,9 +5,11 @@ type t
 val from_string : string -> t
 val to_string : t -> string
 
-val repo_name : t -> string
+val repo_name : t -> (string, Rresult.R.msg) result
 (** Returns the name of the repo given the dev-repo.
     E.g. [repo_name (from_string "https://github.com/ocamllabs/opam-monorepo.git")]
-    returns ["opam-monorepo"]. *)
+    returns ["opam-monorepo"].
```
Leonidas-from-XIV, line 11: Can you describe the "high-level" approach it does here, e.g. determining the last component and removing everything before the dot? I think having a english language specification on how the repo name is determined would be useful.

lib/pull.ml
```diff
@@ -5,17 +5,31 @@ let pull ?(trim_clone = false) ~global_state ~duniverse_dir src_dep =
...
-  let* () = result in
-  if trim_clone then
-    let* () =
-      Bos.OS.Dir.delete ~must_exist:false ~recurse:true
-        Fpath.(output_dir / ".git")
+  if
+    Fpath.(is_prefix (normalize duniverse_dir) (normalize output_dir))
+    && not (String.equal dir "")
+  then
+    let url = Url.to_opam_url url in
```
Leonidas-from-XIV, line 12: Maybe this part should be split into it's own function, since long `if`-blocks followed by rather large `else` blocks are hard to read.

test/lib/test_dev_repo.ml
```diff
@@ -3,17 +3,19 @@ let test_repo_name =
...
+      ~expected:(Ok "opam-monorepo") ();
     make_test ~dev_repo:"git://github.com/ocamllabs/opam-monorepo.git"
-      ~expected:"opam-monorepo" ();
+      ~expected:(Ok "opam-monorepo") ();
     make_test ~dev_repo:"git+https://github.com/ocamllabs/opam-monorepo.git"
-      ~expected:"opam-monorepo" ();
+      ~expected:(Ok "opam-monorepo") ();
+    make_test ~dev_repo:"git+https://github.com/ocamllabs/opam-monorepo/"
+      ~expected:(Ok "opam-monorepo") ();
   ]
```
Leonidas-from-XIV, line 19: Please also add some tests where it fails to parse the dev-repo.

test/lib/test_duniverse.ml
```diff
@@ -192,16 +197,17 @@ module Repo = struct
...
+             {
+               dir = "d";
+               url = Other "u2";
+               hashes = [];
+               provided_packages =
+                 [
+                   opam_factory ~name:"d" ~version:"1";
+                   opam_factory ~name:"d-lwt" ~version:"2";
+                 ];
+             })
```
Leonidas-from-XIV, line 210: Likewise, please add a test where this produces an `Error`

lib/dev_repo.ml
```diff
@@ -8,15 +8,28 @@ let to_string t = t
...
   in
-  match String.lsplit2 ~on:'.' last_path_component with
-  | None -> last_path_component
-  | Some (repo_name, _ext) -> repo_name
+  let name =
+    match String.lsplit2 ~on:'.' last_path_component with
+    | None -> last_path_component
+    | Some (repo_name, _ext) -> repo_name
+  in
+  if String.equal name "" then
```
Leonidas-from-XIV, line 26: I'd just match on `name` here, but that's just me.

```ocaml
match name with
| "" -> Rresult.R.error_msgf "unexpected empty string while computing name for dev_repo: %s" t
| name -> Ok name
```

lib/pull.ml
```diff
@@ -5,17 +5,31 @@ let pull ?(trim_clone = false) ~global_state ~duniverse_dir src_dep =
...
-  let url = Url.to_opam_url url in
-  let open OpamProcess.Job.Op in
-  Opam.pull_tree ~url ~hashes ~dir:output_dir global_state @@| fun result ->
-  let* () = result in
-  if trim_clone then
-    let* () =
-      Bos.OS.Dir.delete ~must_exist:false ~recurse:true
-        Fpath.(output_dir / ".git")
+  if
+    Fpath.(is_prefix (normalize duniverse_dir) (normalize output_dir))
```
Leonidas-from-XIV, line 9: Maybe make this a separate function `is_in_duniverse_dir` or similar, to explain why it is doing this check through naming.

Leonidas-from-XIV (changes requested):
Good catch! It makes a lot of sense to make sure `opam-monorepo pull` doesn't write all over the users disk. I like that the tests also check for directory traversal out of the `duniverse` directory.

I've added some comments where I think things could be clearer, but I think the approach is good.

lib/dev_repo.ml
gridbugs: I ended up rewriting this whole function after I found some more edge cases that it doesn't handle (multiple trailing slashes, leading periods in the final component of the path)

gridbugs:
I've updated the code based on feedback but I haven't written any more tests yet. Will do tomorrow.

lib/pull.ml
```diff
@@ -1,21 +1,40 @@
 open Import
 
+(* Check that [output_dir] is strictly a descendant of [duniverse_dir] *)
+let is_in_universe_dir ~duniverse_dir ~output_dir =
+  Fpath.(is_prefix (normalize duniverse_dir) (normalize output_dir))
+  && not (String.equal (Fpath.filename output_dir) "")
+
+(* Delete version control metadata and vendor sudbirectory *)
```
Leonidas-from-XIV, line 8:
```suggestion
(* Delete version control metadata and vendor subdirectory *)
```

stdext/string.ml
```diff
@@ -42,7 +42,7 @@ let drop_prefix s ~prefix =
 
 let drop_suffix s ~suffix =
   if is_suffix s ~suffix then
-    if length s = length suffix then Some s
+    if length s = length suffix then Some ""
```
Leonidas-from-XIV, line 45: I am wondering whether there was some reason to keep it that (odd) way, since it specifically tests for this case.

After all, if you leave that check out you'd get exactly `Some ""` via `sub s ~pos:0 ~len:0`.

test/bin/pull-invalid-path.t/run.t
```diff
@@ -0,0 +1,18 @@
...
+deleted outside the duniverse directory.
+
+Attempt to pull with a lockfile containing a package whose name is the empty
+string:
+  $ opam-monorepo pull --lockfile=lockfile-refers-to-duniverse-directory.opam.locked
+  ==> Using lockfile lockfile-refers-to-duniverse-directory.opam.locked
+  opam-monorepo: [ERROR] Refusing to pull https://foo.com/foo.tbz into directory $TESTCASE_ROOT/duniverse/ as it is not inside the directory $TESTCASE_ROOT/duniverse
+  [1]
+
+Attempt to pull with a lockfile containinga package whose name is "..":
```
Leonidas-from-XIV, line 14:
```suggestion
Attempt to pull with a lockfile containing a package whose name is "..":
```

test/lib/test_dev_repo.ml
```diff
@@ -3,17 +3,44 @@ let test_repo_name =
...
+          [
+            "";
+            ".";
+            "/";
+            "./.";
+            "...";
+            "///";
+            "https://github.com";
+            "https://github.com/";
+          ]
```
Leonidas-from-XIV, line 33: :+1:

test/lib/test_stdext.ml
```diff
@@ -0,0 +1,47 @@
...
+      make_test "foo" ~suffix:"" ~expected:(Some "foo");
+      make_test "" ~suffix:"foo" ~expected:None;
+      make_test "foo" ~suffix:"o" ~expected:(Some "fo");
+      make_test "foo" ~suffix:"f" ~expected:None;
+    ]
+
+  let tests = List.concat [ drop_prefix; drop_suffix ]
+end
+
+let suite = ("Stdext", Test_string.tests)
```
Leonidas-from-XIV, line 47: This is nice to detect issues, though maybe we shouldn't go to deep into the alternate stdlib game. It could be better to use `Base.String.val chop_suffix` since we have a direct dependency on `base` anyway. I did a quick check and it looks like it passes the tests exactly.

(I also thought `Astring` which we probably indirectly depend on might have something like that and while it has `Astring.String.is_prefix` exists, I can't find a way to drop the affix)

Leonidas-from-XIV (approved):
This looks solid. I've found a few small typos on the way and also some musing about the surprising (existing) behavior of `drop_suffix` which is puzzling but I'm ok merging it like this.

Thanks for the work :)

stdext/string.ml
gridbugs: That's true, but `drop_prefix` also has that check and it returns `""` if the prefix matches the string. I assumed the check was an optimization. The implementation of `String.drop_suffix` in dune's alternate stdlib also used to return the string when it matched the suffix but I changed it a few months ago to return `""`. The only place where `drop_suffix` is called is in dev_repo.ml, so no other code will be affected by this change. I propose we keep my change as it's now consistent with `drop_prefix` and also I think consistent with how most people will expect it to behave (or use `Base.String` and remove this function altogether).

test/lib/test_stdext.ml
gridbugs: I've switched to using `Base.String.chop_suffix` and `chop_prefix` and removed the now unused `String.drop_suffix`

stdext/string.ml
gridbugs: I've switched to using `Base.String.chop_suffix` and `chop_prefix` and removed the now unused `String.drop_suffix`

Leonidas-from-XIV:
LGTM! :)

PR #360, opened 2022-12-06
Fix sort order of `avoid-versions`

Leonidas-from-XIV:
Previously the solver discarded everything in the back of the list, thus not only avoiding the avoid-version packages but effectively removing them from the candidate set. Now we keep the groups of packages consecutive which makes the solver pick them if there is no other option, while still trying to avoid them if other options are available.


Leonidas-from-XIV:
Closing due to the fix not making a difference and the additional test not providing much value.

PR #365, opened 2022-12-23
Normalize Github URLs before comparing

### LINKED ISSUE #118: Semantically equivalent but different "dev-repo" fields lead to build errors
The packages `mirage-clock` and `mirage-clock-freestanding` are provided by the same repository but their `dev-repo` fields are different: `git://github.com/mirage/mirage-clock` and `git+https://github.com/mirage/mirage-clock.git`. 
This causes duniverse to download the mirage-clock repo twice under different names, leading to a build error further on are some packages are defined twice.

Leonidas-from-XIV:
This PR tries to normalize the `dev-repo` URLs somewhat. @TheLortex reported in #118 that the `.git` is differing between the repos so this PR tries to normalize the URLs:

1. Always have `.git` in the repo path
2. Always use `git+https` as scheme
3. Always strip fragments (No more `#main`)

This is just a workaround since there is no way to make sure the URLs are equivalent without visiting them (especially the `.git` addition is a bit dodgy), but it should at least cover some common cases and in the worst case the canonicalization gives up and returns the input, thus equivalent to what was there before.

Also turns out there was a `Uri_utils.has_git_extension` which I assume is somewhat related to the issue at hand, but it was unused so I removed it. 

Closes #118

### DIFF
--- a/lib/uri_utils.ml
+++ b/lib/uri_utils.ml
@@ -1,16 +1,42 @@
 open Import
 
+let path_ok path = path |> Fpath.to_string |> Result.ok
+let flip f a b = f b a
+
+let dump_result =
+  let dump_msg ppf (`Msg s) = Fmt.pf ppf "`Msg %S" s in
+  Fmt.Dump.result ~ok:Fmt.Dump.string ~error:dump_msg
+
 let canonicalize uri =
   let open Result.O in
-  let path = Uri.path uri in
   let new_path =
-    let* fpath = Fpath.of_string path in
-    match Fpath.has_ext ".git" fpath with
-    | true -> Ok fpath
-    | false -> Ok (Fpath.add_ext ".git" fpath)
+    let* fpath = Fpath.of_string (Uri.path uri) in
+    let fpath =
+      match Fpath.has_ext ".git" fpath with
+      | true -> fpath
+      | false -> Fpath.add_ext ".git" fpath
+    in
+    path_ok fpath
   in
-  match new_path with
-  | Error _ -> uri
-  | Ok new_path ->
-      let new_path = Fpath.to_string new_path in
-      Uri.with_path uri new_path
+  let new_scheme =
+    match Uri.scheme uri with
+    | Some "https" -> Ok "git+https"
+    | Some "git" -> Ok "git+https"
+    | Some ("git+https" as git_https) -> Ok git_https
+    | Some other_scheme ->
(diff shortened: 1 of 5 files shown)


lib/uri_utils.ml
```diff
@@ -1,9 +1,42 @@
...
   let open Result.O in
-  let ext_res =
-    let+ path = Fpath.of_string (Uri.path uri) in
-    Fpath.get_ext ~multi:true path
+  let new_path =
+    let* fpath = Fpath.of_string (Uri.path uri) in
+    let fpath =
+      match Fpath.has_ext ".git" fpath with
+      | true -> fpath
+      | false -> Fpath.add_ext ".git" fpath
```
gridbugs, line 17: Is it possible that the uri refers to a non-git repo?

gridbugs (commented):
Does opam-monorepo make the assumption that all dev-repos are git repos?

lib/uri_utils.ml
Leonidas-from-XIV: In theory it can, but

1. Most `dev-repo`s these days are git and even on github.
2. It doesn't matter all that much because this is only to make the URLs canonical for comparison, so `https://hghub.com/foo/bar.hg` would be rewritten to `https+git://hghub.com/foo/bar.hg.git` which might not be a valid URL but will compare equal to itself.

The canonicalization process adds `.git` and rewrites schemes to `git+https` but if you'd rather have it remove `.git` and rewrite `git+https://` and `git://` to `https://` that's also totally ok with me.

Leonidas-from-XIV:
I've decided to wrap the normalization in an abstract type to which the conversion only goes one way, to make sure the canonical URLs (whatever they might be) aren't used for other purposes.

emillon (commented):
Just to be on the safe side, I think it's better to only do this canonicalization for known hosts (for example we know that github routes `user/repo.git` and `user/repo` to the same place but that's not necessarily the case for all hosts. Actually an alternative technique that might be easier (rather than canonicalization) is to parse the dev-repo field into `Github of user:string;repo:string|...|Other of Uri.t`.

Leonidas-from-XIV:
@emillon You're right, that's a good point. I've simplified the whole thing to just parse user/repo out of github.com URLs and keep the rest the same.

It might make sense to remove the hash/branch, but in practice I doubt someone would specify a git repo with a hash as `dev-repo`.

gridbugs (approved):

PR #366, opened 2022-12-27
Preserve symlinks in `duniverse/` if requested

### LINKED ISSUE #348: pull overwrites everything
I use mirage, which uses opam-monorepo. Now I sometimes want to improve one package and test changes within a unikernel setup. How I achieve this is `rm -rf duniverse/my-package ; cd duniverse ; ln -s ../../my-package . ; cd ..` -- so removing the stuff duniverse pulled, and creating a link to my local development version.

Now, when I have 5 packages symlinked, eventually I want to revert some of them to the upstream version. What I attempt to do is `rm duniverse/my-package` - removing the symlink. Now, a `opam monorepo pull` removes _all_ symlinks.

The question I have now: is there a better supported workflow to achieve the same? Is there a way to tell "opam monorepo" to only pull non-existing things, rather than overwriting everything?

Leonidas-from-XIV:
Closes #348.

### DIFF
--- a/lib/pull.ml
+++ b/lib/pull.ml
@@ -62,12 +62,28 @@ let mark_duniverse_content_as_vendored ~duniverse_dir =
   Logs.debug (fun l -> l "Successfully wrote %a" Pp.Styled.path dune_file);
   Ok ()
 
-let pre_pull_clean_up ~full ~duniverse_dir duniverse =
-  if full then Bos.OS.Dir.delete ~must_exist:false ~recurse:true duniverse_dir
-  else
-    Result.List.iter duniverse ~f:(fun { Duniverse.Repo.dir; _ } ->
-        Bos.OS.Dir.delete ~must_exist:false ~recurse:true
-          Fpath.(duniverse_dir / dir))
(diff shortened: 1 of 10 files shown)


lib/pull.ml
```diff
@@ -109,12 +125,24 @@ let write_duniverse_dir_documentation ~duniverse_dir =
   in
   written
 
-let duniverse ~full ~root ~global_state ~trim_clone duniverse =
+let filter_preserved ~preserved duniverse =
+  List.filter_map
```
gridbugs, line 129: I think this could be simplified to
```ocaml
List.filter ~f:(fun entry : Duniverse.Repo.t -> not (List.mem entry.dir ~set:preserved)) duniverse
```
But I get if you think it's easier to read in its current form.

Leonidas-from-XIV: True, good catch. There is a point about readability but there's another point about not overcomplicating it and I did this accidentally. Thanks for pointing it out, I missed the forest for the trees.

gridbugs (approved):

PR #367, opened 2022-12-30
WIP Rewrite unused package names

Leonidas-from-XIV:


lib/dune_file.ml
```diff
@@ -67,6 +67,143 @@ module Lang = struct
...
+  open Sexplib0.Sexp
+  module Set = Set.Make (String)
+  module Map = Map.Make (String)
+
+  type t = Random.State.t
+
+  let init = Random.State.make_self_init
+
+  let random_valid_identifier t =
+    let int_range ~start ~end_ = Random.State.int t (end_ - start) + start in
```
gridbugs, line 80: What's the benefit of using a random identifier over an incrementing counter?

lib/dune_file.ml
```diff
@@ -67,6 +67,143 @@ module Lang = struct
...
+      List.filter_map stanzas ~f:(function
+        | List [ Atom candidate; Atom value ] -> (
+            match String.equal candidate name with
+            | true -> Some value
+            | false -> None)
+        | _ -> None)
+    in
+    match matches with [] -> None | [ x ] -> Some x | _ -> None
+
+  (* type rename_result = { *)
```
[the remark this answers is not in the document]
Leonidas-from-XIV: I've considered using this type instead of a somewhat obscure 3-tuple as return value. Not sure if it makes things clearer, maybe.

lib/dune_file.ml
Leonidas-from-XIV: Indeed the counter would be simpler, but also more likely to lead to duplicated identical names, given the way the renaming is implemented:

Currently the renaming is done per directory, which is also where the random state is initialized, so each vendored, unpacked tarball gets a new state. If I changed it to use a counter, two tarballs both having an identifier called `bar` that needs renaming and using `%s.a%d` as renaming scheme would end up being called `bar.a0`, thus creating a conflict.

Of course the code could be changed to initialize the deduplication counter once per deduplication run and then have all tarballs use the same counter but that I feel keeping it more separated makes it slightly nicer, since it shares less state overall and instead relies more on the randomness of 8-ASCII characters.

If you have strong opinions on that or good arguments for using a counter that I have missed, I'll gladly revise this code.

emillon: I'm not sure I follow why there's a need for a random ID, but I think it can be an issue to add some nondeterminism in the `lock` step. Is there a way to use a deterministic function for this, like a hash of package names in collision or something like that?

lib/dune_file.ml
```diff
@@ -67,6 +67,143 @@ module Lang = struct
...
+  (* determine whether the package should be kept or not, handles [pkg.name] and [pkg] *)
+  let should_keep ~keep name =
+    let test_against =
+      match Base.String.lsplit2 name ~on:'.' with
+      | Some (package, _name) -> package
+      | None -> name
+    in
+    Set.mem test_against keep
+
+  let rename_library t ~keep renames stanzas =
```
emillon, line 122: One trick I recently learned is that you can patch the `dune` file of a subdirectory without writing in that directory: you can mark that directory as `data_only_dirs` and use `(subdir)` to set the contents of the (virtual) dune file in there. I don't know if that applies to that case, but that can be useful if you don't want to change the contents of the unpacked tarball.

lib/dune_file.ml
Leonidas-from-XIV: The non-determinism is in the pulling not the locking but indeed you're right, being able to derive the distinction value from a hash of the folder names (which have to be unique already) works better and allows for a deterministic renaming strategy.

I've applied a change to use `Digest.string` and it made the code a bit simpler even.

lib/pull.ml
```diff
@@ -14,17 +14,93 @@ let do_trim_clone output_dir =
...
+  | None -> Ok None
+  | Some sexps -> (
+      let Dune_file.Packages.{ changed; stanzas; renames } =
+        Dune_file.Packages.rename dfp ~keep renames sexps
+      in
+      match changed with
+      | false -> Ok None
+      | true -> Ok (Some (stanzas, renames)))
+
+(* can't use Sexplib pretty printer because it tries to hard to escape values like \ *)
```
gridbugs, line 37: "...tries _too_ hard..."

gridbugs (approved):

PR #368, opened 2023-01-10
Update the lockfile to OCaml 5.0

tmattio:


Leonidas-from-XIV:
@tmattio Can you please rebase on `main`? I've updated the repo to use ocamlformat that works with OCaml 5.0.

Leonidas-from-XIV:
I guess then we should be ok to merge :+1:

PR #372, opened 2023-02-03
Do not rewrite `dune` files in ML format when updating references

Leonidas-from-XIV:
Currently `opam-monorepo` is skipping over files with the Tuareg stanza when renaming, but not when updating the references. Oddly enough Sexplib parses these files successfully, but outputting them breaks down so we need to make sure to skip files with the Tuareg stanza even when updating the references.

### DIFF
--- a/lib/dune_file.ml
+++ b/lib/dune_file.ml
@@ -169,8 +169,11 @@ module Packages = struct
         | None -> { changed = false; stanzas = original; renames })
     | otherwise -> { changed = false; stanzas = otherwise; renames }
 
+  (* the file is an ML file if the first stanza starts with the emacs tuareg mode
+     comment *)
   let is_tuareg = function
-    | List (Atom "*" :: Atom "-*-" :: Atom "tuareg" :: Atom "-*-" :: _) -> true
+    | List (Atom "*" :: Atom "-*-" :: Atom "tuareg" :: Atom "-*-" :: _) :: _ ->
+        true
     | _ -> false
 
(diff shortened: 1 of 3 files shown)


lib/dune_file.ml
```diff
@@ -198,26 +202,32 @@ module Packages = struct
...
-            update_reference renames sexp
-          in
-          (changed_before || changed, stanzas :: acc))
-        ~init:(false, []) sexps
-    in
-    let sexps = List.rev sexps in
-    { changed; stanzas = sexps; renames }
+    let changed = false in
+    match sexps with
+    | [] -> { changed; stanzas = []; renames }
```
gridbugs, line 207: I think the empty list case here is redundant as it's handled by the fold in the non-tuareg case below

lib/dune_file.ml
```diff
@@ -169,8 +169,12 @@ module Packages = struct
         | None -> { changed = false; stanzas = original; renames })
     | otherwise -> { changed = false; stanzas = otherwise; renames }
 
+  let is_tuareg = function
```
gridbugs, line 172: it might be clearer if this function took a list of sexps rather than a single sexp. Tuareg-ness is a property of a dune file (which here is represented as a list of sexps) - not a property of an individual sexp, as it would be meaningless to call this function on (say) the second sexp in the file. I think making this change will also remove the need for explicit pattern matching in the places where this function is used.

gridbugs (approved):

PR #376, opened 2023-02-20
Mark packages as virtual when they have neither `build` nor `install`

Leonidas-from-XIV:
Reported by @gridbugs that `xml-light.2.4` is missing `build` but is not a virtual package due to having `install`.

This simplifies the previous detection a little bit, since the build commands (and by extension the install commands) are never used, so just recording their presence or absence should be enough.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -27,6 +27,10 @@
 - Fix assertion failure when prefix of "lock" subcommand is used (#381,
   @gridbugs)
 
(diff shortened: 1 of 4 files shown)


gridbugs (approved):

gridbugs:
Sounds good. Sorry this slipped off my plate.

Leonidas-from-XIV:
Cheers; I also forgot it and only noticed again when going through the list of PR as I was considering cutting a new point release.

PR #377, opened 2023-02-21
Error when multiple dev-repos collide on dir

gridbugs:
The duniverse directory of a package is implied by its dev-repo. It's possible for different dev-repos to map to the same directory. Prior to this change, one lucky package would be downloaded to the expected place and the others would be silently ignored. More precisely, each package would be downloaded and placed in the duniverse dir implied by their dev-repo, but first any existing duniverse dir of the same name would be removed.

After this change, the same situation results in an error when running `opam monorepo lock`.

### DIFF
--- a/lib/dev_repo.ml
+++ b/lib/dev_repo.ml
@@ -5,7 +5,7 @@ type t = string
 let compare = String.compare
 let from_string s = s
 let to_string t = t
-let pp fmt t = Format.fprintf fmt "%s" t
+let pp = Fmt.string
 
 let rec repeat_while_some x ~f =
   match f x with None -> x | Some x -> repeat_while_some x ~f
--- a/lib/dev_repo.ml
+++ b/lib/dev_repo.ml
@@ -5,6 +5,7 @@ type t = string
 let compare = String.compare
 let from_string s = s
 let to_string t = t
+let pp fmt t = Format.fprintf fmt "%s" t
 
 let rec repeat_while_some x ~f =
   match f x with None -> x | Some x -> repeat_while_some x ~f
--- a/lib/duniverse.ml
+++ b/lib/duniverse.ml
@@ -231,6 +231,58 @@ let dev_repo_map_from_packages packages =
         | Some pkgs -> Some (pkg :: pkgs)
         | None -> Some [ pkg ]))
 
+(* converts a map from dev-repos to lists of packages to a list of repos after
+   checking for errors *)
+let dev_repo_package_map_to_repos dev_repo_package_map =
+  let open Result.O in
+  let dev_repo_to_repo_result_map =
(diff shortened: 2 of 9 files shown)


lib/dev_repo.ml
```diff
@@ -5,6 +5,7 @@ type t = string
 let compare = String.compare
 let from_string s = s
 let to_string t = t
+let pp fmt t = Format.fprintf fmt "%s" t
```
Leonidas-from-XIV, line 8:
```suggestion
let pp = Fmt.string
```

lib/duniverse.ml
```diff
@@ -231,6 +231,58 @@ let dev_repo_map_from_packages packages =
...
+    in
+    match
+      List.find_opt dev_repos_by_dir ~f:(fun (_, dev_repos) ->
+          List.length dev_repos > 1)
+    with
+    | None -> Ok ()
+    | Some (dir, dev_repos) ->
+        let dir_path = Fpath.(Config.vendor_dir / dir) in
+        let message_first_line =
+          Format.asprintf
```
Leonidas-from-XIV, line 269:
```suggestion
          Fmt.str
```

lib/duniverse.ml
```diff
@@ -231,6 +231,58 @@ let dev_repo_map_from_packages packages =
...
+            "Multiple dev-repos would be vendored into the directory: %a"
+            Fpath.pp dir_path
+        in
+        let message_dev_repos =
+          Format.sprintf "Dev-repos:\n%s"
+            (List.map dev_repos ~f:(fun dev_repo ->
+                 Format.asprintf "- %a" Dev_repo.pp dev_repo)
+            |> String.concat ~sep:"\n")
```
Leonidas-from-XIV, line 277: This is nicer solved by using `Fmt.list` which can format list of formatters, thus solving the issue of concatenating and mapping, so you only need a formatter that formats a `dev_repo` in the right way (`fun ppf dev_repo -> Fmt.pf ppf "- %a" Dev_repo_pp dev_repo`).

lib/duniverse.ml
```diff
@@ -231,6 +231,58 @@ let dev_repo_map_from_packages packages =
...
+        in
+        let message =
+          [ message_first_line; message_dev_repos ] |> String.concat ~sep:"\n"
+        in
+        Error (`Msg message)
```
Leonidas-from-XIV, line 282: I'd use `Rresult.R.error_msgf "%s\n%a" message_first_line pp_dev_repos message_dev_repos` or something like this here.

Leonidas-from-XIV (commented):
Looks good and a useful addition to prevent confusion, the only suggestions I have is to replace the string construction with formatters.

gridbugs:
Thanks for the feedback. It's much more concise now!

PR #381, opened 2023-04-13
Fix crash when prefix of "lock" subcommand is used

gridbugs:


cli/lock.ml
```diff
@@ -515,7 +515,7 @@ let extract_source_config ~adjustment ~opam_monorepo_cwd ~opam_files
 
 let raw_cli_args () =
   match Array.to_list Sys.argv with
-  | _bin :: "lock" :: args -> args
+  | _bin :: prefix :: args when String.starts_with ~prefix "lock" -> args
```
Leonidas-from-XIV, line 518: I think this would lead to a weird case where `opam-monorepo locksmith` goes into this case.

I am somewhat confused how you got into this place, since I would've expected `opam-monorepo locksmith` not resolving to this module to start with.

Leonidas-from-XIV (commented):
Could you explain how to get to this error? I can't figure out how this would trigger (the original code uses `assert false` probably for this exact reason - the code is believed to be unreadable).

gridbugs:
> Could you explain how to get to this error? I can't figure out how this would trigger (the original code uses `assert false` probably for this exact reason - the code is believed to be unreadable).

You can trigger this by running `opam monorepo loc`. The output looks like it will work but then it crashes right at the end (presumably when it tries to add the command to the lockfile). For example:
```
$ opam monorepo loc
opam-monorepo: [WARNING] Package x is defined multiple times in the repository:
- /home/s/src/opam-monorepo/tmp/x.opam
- /home/s/src/opam-monorepo/test/bin/error-on-directory-conflict.t/x.opam
We kept /home/s/src/opam-monorepo/test/bin/error-on-directory-conflict.t/x.opam and discarded the others
[WARNING] Unknown variable "ocaml-system:version"
[WARNING] Unknown variable "ocaml-base-compiler:version"
[WARNING] Unknown variable "ocaml-variants:version"
==> Found 53 opam dependencies for the target package.
               File "cli/lock.ml", line 519, characters 9-15: Assertion failed
               Raised at Duniverse_cli__Lock.raw_cli_args in file "cli/lock.ml", line 519, characters 9-21
               Called from Duniverse_cli__Lock.run in file "cli/lock.ml", line 554, characters 17-32

```

cli/lock.ml
gridbugs: It will only match prefixes of "lock", not strings which "lock" is a prefix of. `cmdliner` accepts prefixes of subcommands already. This code here is a sanity check that the command was a `opam monorepo lock` subcommand when saving the command to the lockfile.

Leonidas-from-XIV: Ah, you're right, I've misread the code. Honestly, I think the previous code is fine and we/Cmdliner should not accept prefix matches. If the user uses `opam monorepo l`  and we introduce a new command that starts with `l` then suddenly things that worked don't anymore because it is impossible to disambiguate. Which can be annoying in e.g. scripts like the CI.

Leonidas-from-XIV:
It seems like the prefix matching of Cmdliner can't be disabled at the moment, so I created a ticket: https://github.com/dbuenzli/cmdliner/issues/178

Leonidas-from-XIV (approved):
I'm ok with merging this if you think that this fix is useful in the meantime, the code itself is fine.

gridbugs:
I think I agree that I'd rather not allow any prefixes for subcommands. Until we can totally disable prefixes in cmdliner I think this change is still valuable as it's currently confusing that `opam monorepo lo` fails in the way it does (it should either immediately fail due to an invalid subcommand or completely work), especially when all other subcommands can be prefixed.

Leonidas-from-XIV:
Turns out the change in Cmdliner was refused, so we'd have to do our own code to disable this behavior :-(

PR #382, opened 2023-04-14
Fix for Dockerfile smell DL3059

grosa1:
Hi!
The Dockerfile placed at "Dockerfile.test" contains the best practice violation [DL3059](https://github.com/hadolint/hadolint/wiki/DL3059) detected by the [hadolint](https://github.com/hadolint/hadolint) tool.

The smell DL3059 occurs if there are multiple consecutive RUN instructions. Each RUN will correspond to a layer of the final Docker image, thus is recommended to compact them to reduce the number of layers for the final image.
This pull request proposes a fix for that smell generated by my fixing tool. The patch was manually verified before opening the pull request. To fix this smell, specifically, The consecutive RUN instructions have been chained until a comment or a different instruction appears.

This change is only aimed at fixing that specific smell. If the fix is not valid or useful, please briefly indicate the reason and suggestions for possible improvements.

Thanks in advance

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -4,8 +4,8 @@ RUN opam install -y dune
 COPY --chown=opam . /home/opam/src/
 WORKDIR /home/opam/src
 ENV DUNE_CACHE=enabled
-RUN --mount=type=cache,target=/home/opam/.cache,uid=1000 opam exec -- dune build --profile=static @install
-RUN --mount=type=cache,target=/home/opam/.cache,uid=1000 opam exec -- dune install
+RUN --mount=type=cache,target=/home/opam/.cache,uid=1000 opam exec -- dune build --profile=static @install \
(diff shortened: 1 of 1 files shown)


Leonidas-from-XIV:
Well that particular patch won't work because the syntax is wrong. `--mount=type=cache,target=/home/opam/.cache,uid=1000` are arguments to `RUN`, not commands to be executed so if contracting the commands via `&&` they'd need to be remove.

grosa1:
Oh well, good point. This is a false positive that should be excluded.
Thank you for the response!

Leonidas-from-XIV:
It can probably still be improved by joining those but given the file is only for testing purposes the additional layer of caching could be useful.

OTOH, this dockerfile is so old and outdated, it still uses the old name of the binary and doesn't work anyway. Maybe we should just remove it instead.

PR #387, opened 2023-04-28
Use `is_prefix` which is available in OCaml 4.10

Leonidas-from-XIV:
#381 introduced an additional check, but unfortunately `String.starts_with` was only added in OCaml 4.13. This PR uses an equivalent function from Base that is available in OCaml 4.10.


gridbugs (approved):

Leonidas-from-XIV:
It's a bit sad the OCaml-CI either runs only a locked monorepo build but not the regular one in addition, which would make it easier to spot minor issues like this. Maybe worth raising an issue.

PR #394, opened 2023-08-27
Add solver tests

samoht:
I wanted to fix a few annoying UX issues (related to ignoring version constraints in multi-packages repositories) but we need solver tests first.


Leonidas-from-XIV (approved):
I am not sure the version unification problem can be solved as I had to admit defeat in #375, but alternative ideas on solving this are welcome as is this test for the solver. Thanks.

PR #395, opened 2023-08-28
Improve demote - move broken packages to the end

samoht:

Issue #331, opened 2022-08-25
opam monorepo ignores upper bounds


hannesm:
Dear Madam or Sir,

from the following (generated) opam file, the "opam monorepo lock pull" and "dune build" fails. It seems to me that while "hpack 0.9.0" could be used, "h2 0.8.0" must be used (there's an upper bound on "paf 0.9.0" in opam-repository to use "h2 < 0.9.0". Somehow, opam monorepo does not respect this bound -- though pin-depends contains h2 0.8.0, the "duniverse-dirs" only lists "ocaml-h2" with 0.9.0.

Here's the opam file:
```
opam-version: "2.0"
maintainer: "dummy"
authors: "dummy"
homepage: "dummy"
bug-reports: "dummy"
dev-repo: "git://dummy"
synopsis: "Unikernel minipaf - switch dependencies"
description: """
It assumes that local dependencies are already
fetched.
"""

build: ["sh" "-exc" "mirage build"]

install: [
  [ "cp" "dist/minipaf" "%{bin}%/minipaf" ]
]

depends: [
  "dns-client" { ?monorepo & >= "6.2.0" & < "7.0.0" }
  "lwt" { ?monorepo }
  "mirage" { build & >= "4.2.0" & < "4.3.0" }
  "mirage-bootvar-unix" { ?monorepo & >= "0.1.0" & < "0.2.0" }
  "mirage-clock-unix" { ?monorepo & >= "3.0.0" & >= "4.1.0" & < "5.0.0" }
  "mirage-console-unix" { ?monorepo & >= "5.1.0" & < "6.0.0" }
  "mirage-crypto-rng-mirage" { ?monorepo & >= "0.8.0" & < "0.11.0" }
  "mirage-logs" { ?monorepo & >= "1.2.0" & < "2.0.0" }
  "mirage-random" { ?monorepo & >= "3.0.0" & < "4.0.0" }
  "mirage-runtime" { ?monorepo & >= "4.2.0" & < "4.3.0" }
  "mirage-time" { ?monorepo }
  "mirage-unix" { ?monorepo & >= "5.0.0" & < "6.0.0" }
  "ocaml" { build & >= "4.08.0" }
  "opam-monorepo" { build & >= "0.3.2" }
  "paf" { ?monorepo & >= "0.0.9" }
  "tcpip" { ?monorepo & >= "7.0.0" & < "8.0.0" }
]

[...]

PR #396, opened 2023-08-28
Remove custom deduplication of packages and use opam solver instead

samoht:
Convert the constraint "if X and Y are from the same repository" into a conflict relationship in the opam metadata between `X` and all the version of `Y` which do not have the same exact version as `X`.

This allow to remove the adhoc post-processing of opam solution which was going very wrong in lots of case. Hopefully, the results will now be much closer to what you would expect `opam` to produce.

### DIFF
--- a/lib/duniverse.ml
+++ b/lib/duniverse.ml
@@ -122,25 +122,6 @@ module Repo = struct
     provided_packages : OpamPackage.t list;
   }
 
-  let log_url_selection ~dev_repo ~packages ~highest_version_package =
-    let url_to_string : unresolved Url.t -> string = function
-      | Git { repo; ref } -> Printf.sprintf "%s#%s" repo ref
-      | Other s -> s
-    in
-    let pp_package fmt { Package.opam = { name; version }; url; _ } =
-      Format.fprintf fmt "%a.%a: %s" Opam.Pp.package_name name Opam.Pp.version
-        version (url_to_string url)
-    in
-    let sep fmt () = Format.fprintf fmt "\n" in
-    Logs.warn (fun l ->
-        l
-          "The following packages come from the same repository %s but are \
-           associated with different URLs:\n\
-           %a\n\
-           The url for the highest versioned package was selected: %a"
-          (Dev_repo.to_string dev_repo)
-          (Fmt.list ~sep pp_package) packages pp_package highest_version_package)
-
   module Unresolved_url_map = Map.Make (struct
     type t = unresolved Url.t
 
@@ -165,23 +146,18 @@ module Repo = struct
     match urls with
     | [ (url, hashes) ] -> Ok { dir; url; hashes; provided_packages }
     | _ ->
-        (* If packages from the same repo were resolved to different URLs, we need to pick
-           a single one. Here we decided to go with the one associated with the package
-           that has the higher version. We need a better long term solution as this won't
-           play nicely with pins for instance.
-           The best solution here would be to use source trimming, so we can pull each individual
-           package to its own directory and strip out all the unrelated source code but we would
-           need dune to provide that feature. *)
-        let* highest_version_package =
-          Base.List.max_elt packages ~compare:(fun p p' ->
-              OpamPackage.Version.compare p.Package.opam.version p'.opam.version)
-          |> Base.Result.of_option
-               ~error:(Rresult.R.msg "No packages to compare, internal failure")
-        in
-        log_url_selection ~dev_repo ~packages ~highest_version_package;
-        let url = highest_version_package.url in
-        let hashes = highest_version_package.hashes in
-        Ok { dir; url; hashes; provided_packages }
+        let pp_hash = Fmt.of_to_string OpamHash.to_string in
+        (* this should not happen because we passed extra constraints
+           to the opam solver to avoid this situation *)
+        Fmt.failwith
+          "The following packages have the same `dev-repo' but are using  \
+           different versions of the archive tarballs:\n\
+           %a\n\
+           This should not happen, please report the issue to \
+           https://github.com/tarides/opam-monorepo.\n\
+           %!"
+          Fmt.Dump.(list (pair (Url.pp string) (list pp_hash)))
+          urls
 
   let equal equal_ref t t' =
     let { dir; url; hashes; provided_packages } = t in
--- a/lib/opam_solve.ml
+++ b/lib/opam_solve.ml
@@ -199,6 +199,83 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
     | Error _ -> false
     | Ok opam_file -> Opam.has_cross_compile_tag opam_file
 
+  (* this is a hack - as we don't have a global view of the package
+     universe, we build the dev_repo table lazily, whenever the 0install
+     solver request more information about a package (by calling
+     [Solver.candidates]. *)
+  let dev_repos = Dev_repo.Tbl.create 13
+
+  let hashes pkg =
+    match OpamFile.OPAM.url pkg with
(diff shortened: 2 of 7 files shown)


samoht:
No this PR will say there is no solution (which is the fine answer to give here) as foo.1.0 and foo-lwt.1.1 will be in conflict

samoht:
Could you point out to a place where such constraints happen in practice?

samoht:
(Also I'm not sure I understand your example: what bar and bar-lwt have anything to do with foo?)

Leonidas-from-XIV:
> (Also I'm not sure I understand your example: what bar has anything to do with foo?)

Sorry, the example I wrote said `foo` and `bar` but then I realized that `foo` and `foo-lwt` are more real-life examples and forgot to replace `bar` everywhere. I've updated it, could you re-read it and let me know if it is clearer now?

But if now it will fail to solve, this is:

  1. Not what OPAM would do, where `foo.1.1` and `foo-lwt.1.0` would be a valid and compiling solution
  2. Can be achieved easier by just detecting the case of duplicates and instead of deduplicating (and printing a warning about the deduplication) just fail.

The deduplication was added to make solutions that work in OPAM work in opam-monorepo while sidestepping the issues of having the same library multiple times in the duniverse.

#331 is the case you've seen that would be affected. Where `h2` and `hpack` are from the same package (and opam-monorepo requires them to be the same version) whereas a different dependency, `paf` requires a a lower version of `h2` whereas the user code requires a higher version of `h2`. So this is a more complicated example (since the lower-bound is introduced through an external package and as a software author controlling the bounds of packages you depend on is hard).

Thus introducing a conflict artificially would make it fail in the solver already, thus potentially making opam-monorepo fairly useless since a lot of compiling solutions (that work with OPAM) would be rejected.

samoht:
I'm still not sure to understand - this patch does exactly what @reynir is suggesting in https://github.com/tarides/opam-monorepo/issues/331#issuecomment-1477772291: it makes the solver do the deduplication by always generate a consistent solution (for opam-monorepo). I'm ok to diverge from `opam` here slightly: we have extra constraints (that all archives we download are co-installable), so it's better to fail cleanly instead of trying to patch up things randomly. Also: post-processing patching of solutions is generally a bad idea as it might ignore some real constraints.

Leonidas-from-XIV:
But the problem is not a solver problem. If you remove the deduplication the solver will leave you with a perfectly fine solution, the issue will be that this is not a solution that dune can build. The problem has actually very little to do with opam-monorepo, on the other hand, opam-monorepo tries to sidestep it as much as possible by hopefully only fetching tarballs where the `public_name` fields are all unique so dune knows how to build them.

To not split the discussion too much between here and there, I've replied to @reynir's comment on the original issue: https://github.com/tarides/opam-monorepo/issues/331#issuecomment-1697491454 to explain the exact issue with `h2` and `hpack` as specific examples.

Leonidas-from-XIV:
> I'm ok to diverge from opam here slightly. We have extra constraints: all archives we download have to be co-installable, so it's better to fail cleanly with a clear error message instead of trying to patch up things randomly.

Well that's fair, but then we trade the bug that constraints in some cases are ignored (and we try to optimistically use the newer version) with a strict failure in that case (so instead of maybe failing to build it will definitely fail to lock, thus decreasing the amount of packages that can be installed this way).

We can of course do that and I am happy to have a more reliable solution but I also think that users will not be happy if a new version of opam-monorepo can't build a duniverse anymore after an older version was able to make it work (albeit in a rather wonky way).

samoht:
> We can of course do that and I am happy to have a more reliable solution but I also think that users will not be happy if a new version of opam-monorepo can't build a duniverse anymore after an older version was able to make it work (albeit in a rather wonky way).

I am really curious if this is a real issue in practice. I'm really much more inclined to make the tool fail in an understandable way and let packages "fix" potential issues (like the temporary incompatibility of h2 and hpack which was fixed as soon as it was identified properly).

I also think we could make the conflict message more explicit in case the conflict was added because of using the same `dev-repo` but that could be done later on.

I think we should get that one merged and released and see how that works in practice - happy to revert if something terrible happen ;-)

lib/duniverse.ml
```diff
@@ -165,23 +146,13 @@ module Repo = struct
...
-              OpamPackage.Version.compare p.Package.opam.version p'.opam.version)
-          |> Base.Result.of_option
-               ~error:(Rresult.R.msg "No packages to compare, internal failure")
-        in
-        log_url_selection ~dev_repo ~packages ~highest_version_package;
-        let url = highest_version_package.url in
-        let hashes = highest_version_package.hashes in
-        Ok { dir; url; hashes; provided_packages }
+        let pp_hash = Fmt.of_to_string OpamHash.to_string in
+        Fmt.epr "XXX %a\n%!"
```
Leonidas-from-XIV, line 150: I think it would be better to rework that logic under the assumption that there is only one such binding, anything else is a programming error.

lib/opam_solve.ml
```diff
@@ -196,6 +196,86 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
     | Error _ -> false
     | Ok opam_file -> Opam.has_cross_compile_tag opam_file
 
+  (* this is a hack - as we don't have a global view of the package
+     universe, we build the dev_repo table lazily, whenever the 0install
+     solver request more information about a package (by calling
+     [Solver.candidates]. *)
+  let dev_repos = ref Dev_repo.Map.empty
```
Leonidas-from-XIV, line 203: Could also use a `Hashtbl` here? No strong opinions on this either way.

lib/opam_solve.ml
```diff
@@ -196,6 +196,86 @@ module Opam_monorepo_context (Base_context : BASE_CONTEXT) :
...
+  let hashes pkg =
+    match OpamFile.OPAM.url pkg with
+    | None -> []
+    | Some url ->
+        (* note: pinned packages do not have any checksums set *)
+        OpamFile.URL.checksum url
+
+  (* build the list of conflicts by removing packages with the same hash *)
+  let conflicts_with_same_dev_repo_but_a_different_hash =
+    let memo = Hashtbl.create 13 in
```
Leonidas-from-XIV, line 214: Anything particularly useful about picking 13 as initial size or is this just some number?

Leonidas-from-XIV (commented):
> I am really curious if this is a real issue in practice. I'm really much more inclined to make the tool fail in an understandable way and let packages "fix" potential issues

I am unsure, this logic has been added before my involvement in opam-monorepo. We can certainly give it a try.

> I think we should get that one merged and released and see how that works in practice - happy to revert if something terrible happens ;-)

We could release this as 0.4.0 and see what the user feedback will be. Maybe it will be indeed a better user experience!

lib/opam_solve.ml
samoht: That's a good small prime number to init such structures ;-)

lib/duniverse.ml
samoht: I've reworked the error message (and that XXX is a leftover, shouldn't have been there - thanks for spotting this)

lib/opam_solve.ml
samoht: updated with an hashtable

samoht:
For some reasons the CI is broken and doesn't test anything. I think it comes from #395 -- fixing it now.

samoht:
Rebased

Leonidas-from-XIV:
Thanks. I think we should cut a release sometime soon so users can give this a spin and provide feedback.

samoht:
Sounds good - let me know when it's released and I'll try it on mirage stuff :p

PR #397, opened 2023-09-03
Fix issue with #395

samoht:


Leonidas-from-XIV (approved):
Thanks for spotting, I seem to have overlooked it.

The CI looks like it is failing for different reasons, will ping the CI team about this.

PR #398, opened 2023-09-22
Ignore pinned packages when building dev-repo conflicts

samoht:
This allows to pin a single package for multi-repository. In that case Opam-monorepo will always choose the pinned repository, so adding conflicts is just confusing the solver.

### DIFF
--- a/lib/opam_solve.ml
+++ b/lib/opam_solve.ml
@@ -844,23 +844,24 @@ struct
       =
     let env varname = String.Map.find_opt varname env in
     let pins = OpamPackage.Set.of_list pins in
+    let pkg_of_opam pkg =
+      let name = OpamFile.OPAM.name pkg in
+      let version = OpamFile.OPAM.version pkg in
+      OpamPackage.create name version
+    in
     (* remove pinned packages from the universe -- as that's what's the
        opam solver is doing. *)
-    let pp_pkg ppf pkg =
-      Fmt.pf ppf "%s.%s"
-        (OpamPackage.Name.to_string (OpamFile.OPAM.name pkg))
-        (OpamPackage.Version.to_string (OpamFile.OPAM.version pkg))
-    in
     let pkgs =
-      List.filter pkgs ~f:(fun pkg ->
+      List.filter pkgs ~f:(fun opam ->
+          let pkg = pkg_of_opam opam in
           let keep =
             match
-              OpamPackage.package_of_name_opt pins (OpamFile.OPAM.name pkg)
+              OpamPackage.package_of_name_opt pins (OpamPackage.name pkg)
             with
             | None -> true
-            | Some pin -> OpamFile.OPAM.version pkg = OpamPackage.version pin
+            | Some pin -> OpamPackage.version pkg = OpamPackage.version pin
           in
-          Logs.debug (fun l -> l "keep %a = %b" pp_pkg pkg keep);
+          Logs.debug (fun l -> l "keep %a = %b" Opam.Pp.package pkg keep);
           keep)
     in
     { pkgs; constraints; test; env; pins }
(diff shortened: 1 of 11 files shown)


test/lib/test_solve.ml
```diff
@@ -204,11 +204,71 @@ depends: ["p1" "p2"]
...
+url {
+  src: "https://p.com/p.tbz"
+  checksum: "sha256=0000000000000000000000000000000000000000000000000000000000000001"
+}
+|};
+        {|
+opam-version: "2.0"
+name: "ocaml-base-compiler"
+version: "3.14"
+|};
```
Leonidas-from-XIV, line 230: Why is `ocaml-base-compiler.3.14` here twice?

test/lib/test_solve.ml
```diff
@@ -204,11 +204,71 @@ depends: ["p1" "p2"]
...
+  let root =
+    OpamFile.OPAM.read_from_string
+      {|
+opam-version: "2.0"
+name: "root"
+version: "0"
+depends: ["p1" "p2"]
+|}
+  in
+  calculate universe root
```
Leonidas-from-XIV, line 261: I must be missing something but how does this test verify anything about pinning?

Leonidas-from-XIV (commented):
Could you describe the change better? The code doesn't make me go "aha, so this is because of pinned repos", would be nice to have a better description of the change and why it fixes the problem.

test/lib/test_solve.ml
samoht: a pinned package in the opam universe is represented as a package with a single version and no hash. I unfortunately don't think there's a better way to detect this.

reynir: It may be of interest to look at what I did in #353.

samoht: I've pushed a better way to find out pinned packages. It's all a bit of a mess, so I'm not happy with that solution.

I agree we probably need to layer something like what @reynir has proposed to ensure the pinned packages specified in the root packages interact correctly with the pinned packages in the switch.

samoht:
I've pushed a new version: a mix between @reynir's #353 and my patch.

We now have a way to detect if a package is locally pinned, present in the pin-depends, or in a local opam file. This is largely untested for now, any help to push this forward is appreciated.

lib/opam_solve.ml
```diff
@@ -817,11 +836,34 @@ module Mock_context :
...
+      =
     let env varname = String.Map.find_opt varname env in
-    { pkgs; constraints; test; env }
+    let pins = OpamPackage.Set.of_list pins in
+    (* remove pinned packages from the universe -- as that's what's the
+       opam solver is doing. *)
+    let pp_pkg ppf pkg =
+      Fmt.pf ppf "%s.%s"
+        (OpamPackage.Name.to_string (OpamFile.OPAM.name pkg))
+        (OpamPackage.Version.to_string (OpamFile.OPAM.version pkg))
```
Leonidas-from-XIV, line 852: I think `let pp_pkg = Fmt.of_to_string OpamPackage.to_string` can take care of this and will format it exactly in the `<name>.<version>` format.

test/lib/test_duniverse.ml
```diff
@@ -126,16 +128,17 @@ module Repo = struct
                     dev_repo = "d";
                     url = Git { repo = "r"; ref = "master" };
                     hashes = [];
+                    pinned = false;
```
Leonidas-from-XIV, line 131: Could you add a test which sets some pins to see it working?

Leonidas-from-XIV (changes requested):
A few small changes would make it better but the overall form seems ok - let's give it a chance!

samoht:
Thanks, I've pushed fixes to address your review. Not totally sure what tests you wanted me to add, so I've just added a very simple one.

Leonidas-from-XIV (approved):

Leonidas-from-XIV:
Looks good, thanks!

PR #400, opened 2023-10-27
Only add root pin-depends in the .opam.locked file

samoht:
This is a tentative fix for https://github.com/mirage/mirage/issues/1476

I guess we could want to add the switch pins to that pin-depends field for reproducibility.


samoht:
This is ready to review.

This is a breaking change:
- currently, `pin-depends` of the locked file contains the complete list of vendored packages. The rationale is to be able to `opam install <pkg>.opam.locked` to get something similar to `opam monorepo pull`. It was a great initial goal. Still, nowadays, with the distinction between switch local and vendored directories (and the `?vendor` variable), I am not sure this is still very consistent.
- so instead, this PR populates `pin-depends` with the list of packages to be pinned in the current switch. The rationale is to make sure `opam install <pkg>.opam.locked` will install the required dependencies for the existing switch. If there is a `x-opam-monorepo-repositories` field, we skip the locally pinned packages, otherwise, we record them in the locked file.

Note that I am still a bit confused with the expected workflow to use when using `x-opam-monorepo-provided`. Maybe we do not need all of this, but unless I'm mistaken, I don't see how support for pinned packages in the switch is working today.

cli/lock.ml
```diff
@@ -170,6 +170,24 @@ let root_pin_depends local_opam_files =
     local_opam_files []
   |> D.Pin_depends.sort_uniq
 
+let get_local_pins source_config =
+  match (source_config : D.Source_opam_config.t) with
+  | { repositories = Some _; _ } ->
+      (* ignore local pins with using x-opam-monorepo-repositories *)
```
Leonidas-from-XIV, line 176: Small typo:

```suggestion
      (* ignore local pins when using x-opam-monorepo-repositories *)
```

cli/lock.ml
```diff
@@ -170,6 +170,24 @@ let root_pin_depends local_opam_files =
...
+  | _ ->
+      OpamGlobalState.with_ `Lock_none @@ fun global_state ->
+      OpamSwitchState.with_ `Lock_none global_state @@ fun switch_state ->
+      OpamPinned.packages switch_state
+      |> OpamPackage.Set.to_seq
+      |> Seq.filter_map (fun pkg ->
+             match OpamSwitchState.url switch_state pkg with
+             | None ->
+                 (* ignore version-pinned packages *)
+                 None
```
Leonidas-from-XIV, line 187: Maybe it would make more sense to make this into a named function `is_pinned` then it would be clearer that checking for an empty URL means a pinned package (or actually, maybe the OPAM libraries should expose this in some more obviously correct way instead of us trying to interpret what OPAM does).

test/bin/pin-depends.t/run.t
```diff
@@ -0,0 +1,66 @@
...
+    "dune"
+    "b"
+    "c"
+  ]
+  pin-depends: ["b.dev" "./duniverse/b"]
+  x-opam-monorepo-opam-provided: ["b"]
+  x-opam-monorepo-opam-repositories: [
+    "file://$OPAM_MONOREPO_CWD/minimal-repo"
+    "file://$OPAM_MONOREPO_CWD/repo"
+  ]
```
Leonidas-from-XIV, line 17: I don't think printing it here makes sense. If you want to have the contents of the `pin-depends.opam` in the `run.t` I would suggest just creating it with `cat > pin-depends.opam <<EOF`, otherwise its just the same info duplicated.

test/bin/pin-depends.t/run.t
```diff
@@ -0,0 +1,66 @@
...
+  ==> Using 1 locally scanned package as the target.
+  ==> Found 10 opam dependencies for the target package.
+  ==> Querying opam database for their metadata and Dune compatibility.
+  ==> Calculating exact pins for each of them.
+  ==> Wrote lockfile with 1 entries to $TESTCASE_ROOT/pin-depends.opam.locked. You can now run opam monorepo pull to fetch their sources.
+
+The lockfile should contain the base packages, dune and our 2 dependencies
+`b` and `c` which should be pulled in the duniverse
+
+  $ cat pin-depends.opam.locked | sed 's|file://.*/pin-depends.t/|file://$LOCAL_PATH/|'
```
Leonidas-from-XIV, line 32: Could you make this print just the relevant subsets of the lockfile? Otherwise the test is full of incidental content of the lockfile that will need to keep getting updated.

Something like `opam show --just-file --raw -fpin-depends ./pin-depends.opam.locked`.

cli/lock.ml
```diff
@@ -570,11 +588,28 @@ let run (`Root root) (`Recurse_opam recurse) (`Build_only build_only)
...
   Common.Logs.app (fun l -> l "Calculating exact pins for each of them.");
   let* duniverse = compute_duniverse ~dependency_entries >>= resolve_ref in
   let target_depexts = target_depexts opam_files target_packages in
+  let* pin_depends = root_pin_depends opam_files in
+  let local_pins = get_local_pins source_config in
+  let pin_depends =
+    List.filter
+      ~f:(fun (p, _) ->
+        match OpamPackage.Map.find_opt p dependency_table with
+        | None -> assert false
```
Leonidas-from-XIV, line 606: In this case `find` can also be used, if that is not a case that we expect to happen.

Leonidas-from-XIV (changes requested):
Yeah, I think this makes sense.

I am not entirely sure what the original reason was for writing all the package in the `pin-depends` field - maybe indeed it was about being able to reproduce the monorepo with just `opam`. That decision precedes my involvement in opam-monorepo.

PR #401, opened 2023-11-07
Add some helpers to make constructing test-cases simpler

Leonidas-from-XIV:


test/lib/test_solve.ml
```diff
@@ -33,215 +33,129 @@ let calculate universe ?(pins = []) root expected =
...
+let dev_repo = quoted "dev-repo"
+
+let url ~src ?checksum () =
+  match checksum with
+  | Some checksum ->
+      Printf.sprintf {|url {
+  src: %S
+  checksum: "sha256=%064d"
+}|} src
+        checksum
```
Leonidas-from-XIV, line 54: Yes, this looks like a bug in ocamlformat (at least the version we use).

samoht (approved):
That's a nice change, thanks!

PR #403, opened 2024-03-18
Do not load the switch state to get the current OCaml version

samoht:


Leonidas-from-XIV:
Nice, could you add a changelog entry? It seems to me that this would be a useful improvement to point out for users.

Leonidas-from-XIV:
Thank you!

Issue #406, opened 2025-01-19
Trying with Mirage, and llikely with the newly released ocaml-solo5 (1.0.0) supporting OCaml 5.x -- and OCaml 5.3 being released


hannesm:
In a `make lock`, I get the following error:

```
opam-monorepo: [ERROR] Some dependencies cannot be built with dune!

opam-monorepo requires that all dependencies use dune as their build system.

These dependencies (possibly transitive) don't use dune as their build system:
- ocaml-compiler
```

Now, I understand there's been some shuffling around with opam package names and ocaml 5.3 - but I'm a bit stuck what opam-monorepo intends to do here. Would it be possible to fix that?

hannesm:
This has as well been reported on discord on the mirage channel. Would any of those knowing more details of `opam-monorepo` be able to take a look into how to deal with OCaml 5.3? Thanks a lot!

PizieDust:
I have this same issue with OCaml 4.14.1

hannesm:
Please have a look in #407. It would be great to get that merged and released ASAP. Of course any "better" fix is very welcome.

The issue I ran into (and @PizieDust as well) is that albatross has specified a bound on ptime >= 1.2.0, which is not available in the dune-opam-overlays repository. But the ocaml-compiler issue shadowed the underlying issue.

Now, of course it would be great if you could as well put the ptime 1.2.0 package into the dune-opam-overlays repository. In the meantime, I as well opened https://github.com/ocaml/opam-repository/pull/27417 as a workaround.

It would also be nice if your code would print not only the first error, but as well other errors if there are some.

Firobe:
Hi all, opam-monorepo is awaiting release with the #407 fix, and ptime 1.2.0 is now in the dune opam-overlays.

hannesm:
Thanks @Firobe. I'm closing this, since 0.4.1 is now released :)

PR #407, opened 2025-02-10
require opam-state < 2.2.0

hannesm:
OCaml 5.3 introduced a new "ocaml-compiler" package. This does not use dune. This is fine. Addresses #406.


reynir:
It would be great if this could get some attention, merged and released.

Firobe:
Looks good to me as well (CI failure is unrelated), thank you! We'll try to cut a release rapidly (and look at the overlays issues as well).

PR #380, opened 2023-03-21
Upgrade to opam 2.2

kit-ty-kate:
Queued on https://github.com/tarides/opam-monorepo/pull/379

PR #408, opened 2025-02-25
Allow opam-2.3 series

palainp:
Dear developpers,
This PR was needed to compile some of my unikernels with opam 2.3.
Best.

### DIFF
--- a/cli/depext.ml
+++ b/cli/depext.ml
@@ -23,7 +23,6 @@ let run (`Root root) (`Lockfile explicit_lockfile) dry_run (`Yes yes) () =
   let depexts = D.Lockfile.depexts lockfile in
   OpamGlobalState.with_ `Lock_none (fun global_state ->
       let env = OpamPackageVar.resolve_global global_state in
-      let config = OpamFile.Config.with_depext false global_state.config in
       let pkgs =
         List.fold_left
           ~f:(fun acc (pkgs, f) ->
@@ -32,7 +31,7 @@ let run (`Root root) (`Lockfile explicit_lockfile) dry_run (`Yes yes) () =
             else acc)
           ~init:OpamSysPkg.Set.empty depexts
       in
-      let* pkgs = available_packages config pkgs in
+      let* pkgs = available_packages global_state.config pkgs in
       match OpamSysPkg.Set.elements pkgs with
       | [] -> Ok ()
       | pkgs_list ->
@@ -43,7 +42,7 @@ let run (`Root root) (`Lockfile explicit_lockfile) dry_run (`Yes yes) () =
           else if should_install ~yes pkgs_str then
             try
               OpamCoreConfig.update ~confirm_level:`unsafe_yes ();
-              OpamSysInteract.install config pkgs;
+              OpamSysInteract.install global_state.config pkgs;
               Ok ()
             with Failure msg -> Error (`Msg msg)
           else Ok ())
--- a/cli/depext.ml
+++ b/cli/depext.ml
@@ -12,8 +12,8 @@ let should_install ~yes pkgs =
         pkgs)
     ~yes
 
-let available_packages pkgs =
-  match OpamSysInteract.packages_status pkgs with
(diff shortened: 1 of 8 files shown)


palainp:
Sorry, I don't know why it fails, in the CI logs I can see:
```
/src: (run (shell "uname -rs && opam exec -- ocaml -version && opam --version"))
Linux 5.15.0-131-generic
The OCaml toplevel, version 5.0.0
2.2.1
```
Can the opam version in CI be 2.3.0?

Firobe:
Hello, thank you for the PR!

The CI spec generator [hardcodes opam version 2.2](https://github.com/ocurrent/ocaml-ci/blob/e224c585f530a36b15df1762140d25eb82beb352/lib/opam_monorepo.ml#L206), however this shouldn't matter:
- opam-monorepo manages its dependencies with (the released version of) opam-monorepo (see [CONTRIBUTING](https://github.com/tarides/opam-monorepo/blob/main/CONTRIBUTING.md))
- all the dependencies (including `opam*-`!) are locked in `opam-monorepo.opam.locked` and both local devs and the CI are expected to pull the dependencies this way instead of using a global switch (because indeed this is how opam-monorepo is released, with all dependencies vendored)
- the above means that the current release of `opam-monorepo` can be installed no matter if you use `opam` 2.2 and 2.3 (same for `opam-format` and `opam-state`).

So I'm curious what caused you to encounter the problem leading to this PR in the first place?

Regardless, to fix the CI in this PR you should be able to change the lock file so that it points to `opam-format` etc. version 2.3, and the CI will use that :)

palainp:
Thank you for your answer, I'll try to reproduce my issue, and will close the PR if I can't :)

Firobe:
> and will close the PR if I can't :)

It would be good in general to bump the opam dependencies to avoid bitrot (see also https://github.com/tarides/opam-monorepo/pull/380) so the PR doesn't necessarily have to be closed, but I'm just surprised this was blocking to begin with :)

Firobe:
Ah I think I understand: you must have pinned the 5.3 PR, which means the dependencies where expected from the switch instead of being vendored (for release we vendor dependencies, strip all constraints, and then revert after the release...).

So the problem should be solved by cutting a release and using that instead of pinning

palainp:
Yes I think you're totally right :)

Firobe:
Would if be OK for you if I take the lead on completing this PR? (updating the lockfile and fixing any arising errors).

I think it's worthwhile to get this merged before cutting a release, if only to remove the annoying set of `[WARNING] Unknown variable "with-dev-setup"` we get nowadays when building unikernels :p

palainp:
Yes of course!

cli/depext.ml
```diff
@@ -23,6 +23,7 @@ let run (`Root root) (`Lockfile explicit_lockfile) dry_run (`Yes yes) () =
   let depexts = D.Lockfile.depexts lockfile in
   OpamGlobalState.with_ `Lock_none (fun global_state ->
       let env = OpamPackageVar.resolve_global global_state in
+      let config = OpamFile.Config.with_depext false global_state.config in
```
palainp, line 26: In #380 the config was only `global_state.config`, so it may be better to stick with that? In fact, all my modifications are already present in #380 so I think the opam bound can be lowered to 2.2 :)

Firobe: I think you're right. Looking at the called code nothing uses the `depext` field of the config anyway, so I'll go ahead and just use the base config.

And you're right about the bound!

Firobe: done in [ad80a13](https://github.com/tarides/opam-monorepo/pull/408/commits/ad80a134c0520e5f1e1a9fc10cf876696ce772db)

Firobe:
The CI fails for reasons unrelated (and we're trying to investigate why...) but I've confirmed the CI is happy locally! So I'll go ahead and merge, so we can cut a release.

Thank you again!

palainp:
Thank you!

Issue #410, opened 2025-04-08
macos: "opam install opam-monorepo" fails


gm0stache:
dear all,

i am currently trying to install the `opam-monorepo` pkg on my mac with these specs:
```text
OS: macOS 15.4.0
Memory: 16 GiB
Architecture: aarch64
```

the error:
```shell
someuser@somelaptop some-repo % opam install opam-monorepo
[NOTE] It seems you have not updated your repositories for a while. Consider updating them with:
       opam update

The following actions will be performed:
=== install 1 package
  ∗ opam-monorepo 0.4.1

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫
⬇ retrieved opam-monorepo.0.4.1  (cached)
[ERROR] The compilation of opam-monorepo.0.4.1 failed at "dune build -p opam-monorepo -j 7 @install".

#=== ERROR while compiling opam-monorepo.0.4.1 ================================#
# context     2.3.0 | macos/arm64 | ocaml.5.3.0 | https://opam.ocaml.org#91d74cddbf72d811a49b9630c97f596a1cff1b01
# path        ~/.opam/default/.opam-switch/build/opam-monorepo.0.4.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p opam-monorepo -j 7 @install
# exit-code   1
# env-file    ~/.opam/log/opam-monorepo-18279-0b9f17.env
# output-file ~/.opam/log/opam-monorepo-18279-0b9f17.out
### output ###
# [...]
# (cd _build/default/duniverse/base/src && /usr/bin/cc -O2 -fno-strict-aliasing -fwrapv -pthread -D_FILE_OFFSET_BITS=64 -O2 -fno-strict-aliasing -fwrapv -pthread -D_LARGEFILE64_SOURCE -mpopcnt -g -I /Users/g/.opam/default/lib/ocaml -I ../compiler-stdlib/src -I ../hash_types/src -I ../shadow-stdlib/src -I ../../sexplib0/src -o exn_stubs.o -c exn_stubs.c)
# clang: error: unsupported option '-mpopcnt' for target 'arm64-apple-darwin24.4.0'
# File "duniverse/base/src/dune", line 13, characters 35-45:
# 13 |  (c_names exn_stubs int_math_stubs hash_stubs am_testing)
#                                         ^^^^^^^^^^
[...]

PR #411, opened 2025-04-11
Drop dependency on base

reynir:
This drops the dependency on base. In #410 (CC @gm0stache) it is reported that opam-monorepo fails to install on Mac OS on ARM. The error message suggests the compilation fails in base. This rewrites all calls to base to use the standard library. Certainly, there are things that could be factored out to functions, but it's mostly straight forward.

@gm0stache: can you try see if this works for you?

### DIFF
--- a/lib/dev_repo.ml
+++ b/lib/dev_repo.ml
@@ -25,7 +25,7 @@ let repo_name t =
   let path = Uri.of_string t |> Uri.path in
   match Re.exec_opt repo_name_regexp path with
   | Some group ->
-    Ok (Re.Group.get group 0)
+    Ok (Re.Group.get group 1)
   | None ->
     Rresult.R.error_msgf
       "unexpected empty string while computing name for dev_repo: \"%s\"" t
--- a/lib/dev_repo.ml
+++ b/lib/dev_repo.ml
@@ -13,7 +13,7 @@ let repo_name_regexp =
     (seq [
         opt (char '/');
         rep (char '.');
-        group (rep (compl [ char '.'; char '/' ]));
+        group (rep1 (compl [ char '.'; char '/' ]));
         opt (seq [
             char '.';
             rep (compl [ char '/' ])]);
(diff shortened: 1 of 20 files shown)


hannesm:
Dear @Firobe, would it be possible to merge and release this fine PR?

Firobe:
Just waiting to hear back from @gm0stache, and then it's looking good to me :)

gm0stache:
wow! thank you so much @reynir !
means the world to me that you took all the efford to solve my issue!
i just executed `opam pin add opam-monorepo git+https://github.com/tarides/opam-monorepo#8e0d3ee532c778f98bdede8a57499658b27665f9` and it worked!
thank you very much again!!!

lib/dev_repo.ml
```diff
@@ -7,25 +7,27 @@ let from_string s = s
...
-  |> try_split_side ~split:(Base.String.lsplit2 ~on:'.') ~side:fst
-  |> function
-  | "" ->
-      Rresult.R.error_msgf
-        "unexpected empty string while computing name for dev_repo: \"%s\"" t
-  | non_empty -> Ok non_empty
+  let path = Uri.of_string t |> Uri.path in
+  match Re.exec_opt repo_name_regexp path with
+  | Some group ->
+    Ok (Re.Group.get group 0)
```
Firobe, line 27: Thanks @reynir for the PR!

It seems this regexp is not equivalent to the previous code, as some of the tests are failing. For example, `repo_name` returns an error on `git://github.com/ocamllabs/opam-monorepo` instead of `Ok "opam-monorepo"`.

Could you revisit it and make sure the tests are passing locally? (unfortunately the CI is not in a working state)

reynir: I fixed the regular expression and also my usage of `Re.Group`. Two tests are failing due to the ordering of packages not being the same. I will look into that as well.

reynir: Now fixed. The fix is a good argument for factoring out the `Base.Result.all` replacement.

reynir:
I think it should be good now. I suggest squashing all the commits :-)

Firobe:
LGTM, thanks a lot!

PR #415, opened 2025-06-16
Avoid generating conflicts with virtual packages from the same repository

shym:
When a repository contains both virtual and non-virtual packages, avoid the conflict that is generated because the virtual package has no `src` while the non-virtual package has one, so their hash list appear superficially incompatible.

The issue came up with `ocaml-unikraft*` packages: `ocaml-unikraft` is a virtual package to depend on one of the possible packages (according to the target architecture), but the same `dev-repo` also contains the actual non-virtual packages `ocaml-unikraft` is there to depend upon. Before this patch, `opam-monorepo` created a conflict between them.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,10 @@
  
 ### Fixed
 
+- Avoid generating conflicts with virtual packages: virtual and non-virtual
+  packages in the same repository are not considered incompatible anymore
+  (#415, @shym)
+
 ### Removed
 
 ### Security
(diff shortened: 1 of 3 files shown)


reynir:
You could also filter packages that don't have a `src`. If there are no archives to download then it can't conflict.

Firobe:
> You could also filter packages that don't have a `src`. If there are no archives to download then it can't conflict.

Or perhaps  directly reuse the `is_virtual` function if appropriate

Firobe:
I tweaked your `is_virtual` function to exactly match the semantics of the existing one: a package must have both `build` and `install` stanzas absent to be considered virtual.

I also fixed one existing test (where conflicts were not generated anymore, which is OK: I made the test packages non-virtual), and added a new test for the case we're interested in.

I am now relatively confident that this change only removes spurious conflicts. If you're OK with my changes, I'll go ahead and merge.

Issue #416, opened 2025-10-01
opam-monorepo lock --add-opam-repositories throws `OpamUrl.Parse_error`


reynir:
```shell
$ opam-monorepo lock --add-opam-repositories='[opam-overlays git+https://github.com/dune-universe/opam-overlays.git]'
opam-monorepo: internal error, uncaught exception:
               OpamUrl.Parse_error("unsupported version control system \027[04mopam-overlays git\027[0m")
               Raised at OpamUrl.parse_error in file "duniverse/opam/src/core/opamUrl.ml", line 44, characters 20-41
               Called from OpamUrl.parse in file "duniverse/opam/src/core/opamUrl.ml", line 138, characters 19-34
               Called from Stdlib__List.map in file "list.ml", line 83, characters 15-19
               Called from Duniverse_lib__Source_opam_config.Opam_repositories_shape.from_repr in file "lib/source_opam_config.ml", line 78, characters 15-46
               Called from Cmdliner_arg.conv.parse in file "duniverse/cmdliner/src/cmdliner_arg.ml", line 32, characters 22-29
               Called from Cmdliner_base.some.parse in file "duniverse/cmdliner/src/cmdliner_base.ml", line 152, characters 22-29
               Called from Cmdliner_arg.parse_opt_value in file "duniverse/cmdliner/src/cmdliner_arg.ml", line 174, characters 38-45
               Called from Cmdliner_arg.opt.convert in file "duniverse/cmdliner/src/cmdliner_arg.ml", line 192, characters 14-41
               Called from Cmdliner_term.app.(fun) in file "duniverse/cmdliner/src/cmdliner_term.ml", line 22, characters 12-19
               Called from Cmdliner_term.app.(fun) in file "duniverse/cmdliner/src/cmdliner_term.ml", line 19, characters 21-30
               Called from Cmdliner_term.app.(fun) in file "duniverse/cmdliner/src/cmdliner_term.ml", line 22, characters 12-19
               Called from Cmdliner_term.app.(fun) in file "duniverse/cmdliner/src/cmdliner_term.ml", line 22, characters 12-19
               Called from Cmdliner_term.app.(fun) in file "duniverse/cmdliner/src/cmdliner_term.ml", line 22, characters 12-19
               Called from Cmdliner_term.app.(fun) in file "duniverse/cmdliner/src/cmdliner_term.ml", line 19, characters 21-30
[...]

PR #417, opened 2025-10-01
Improve CLI doc and error message

Firobe:
Tries to improve the man page and error message to help with https://github.com/tarides/opam-monorepo/issues/416

- when a list is expected, this is now explicitly said, and the full syntax for list is provided
- all CLI flags that expect such a list now give useful example values in the man page


reynir (approved):
Nice!

